PR #1, opened 2016-08-15
Added trueos_style.

Mrt134:
- Created new directory (themes) and added trueos_style folder.
- Edited conf.py to point to the themes directory and use trueos_style
  when building html.

You may want to add the new themes directory to the preprocessor so the conf.py doesn't have to go up a directory to find the theme:
# Add any paths that contain custom themes here, relative to this directory.
html_theme_path = ['../themes']


userguide/conf.py
```diff
@@ -68,7 +68,7 @@
 }
 
 # Add any paths that contain custom themes here, relative to this directory.
-#html_theme_path = []
+html_theme_path = ['../themes']
```
wblock: This looks like it's referring to a theme directory outside of the build directory (preprocessed).

I wonder if that themes directory should go in _static.  They are static files.

Otherwise, we can just have preproc copy the themes directory along with the other directories in the Makefile:

preproc:
        $(PREPROC) -t $(TAG) -o $(PREPROCDIR) -c 'conf.py _static artwork images themes'

Mrt134:
I don't think there will be any issues with using _static. conf.py will need to be updated to reflect the new path, but there shouldn't be anything else to alter.

wblock:
Moved themes into static, changed the reference in conf.py, no other changes needed.  Nice!

PR #20, opened 2017-06-11
Update Services documentation on stopping/starting

davidjb:
This change updates the documentation to correctly indicate the changes
when starting/stopping Services. For instance, the docs previously
mentioned blue icons, when the icon in the current UI is green.

### DIFF
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -66,21 +66,21 @@ Other services default to off until started.
    Control Services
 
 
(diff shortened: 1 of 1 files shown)


wblock:
Committed in a modified form in https://github.com/freenas/freenas-docs/commit/6fd8870f65f9980bdcadb27cfba86a1e34523a4a

Thanks!

PR #21, opened 2017-07-14
Updates regarding SMB shares

anodos325:
In addition to these, enabling NTLMv1 auth is now handled by an checkbox in the UI under Services->SMB. No need to add an aux parameter.

### DIFF
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1538,15 +1538,6 @@ a multiple of _SC_PAGESIZE (typically *4096*) to avoid memory
 fragmentation. This will increase Samba's memory requirements and
 should not be used on systems with limited RAM.
 
-To increase network performance, read the Samba section on socket
-options in the
-`smb.conf manual page
-<https://www.freebsd.org/cgi/man.cgi?query=smb.conf&manpath=FreeBSD+11.0-RELEASE+and+Ports>`_.
-It indicates which options are available and recommends that you
-experiment to see which are supported by your clients and improve your
-network's performance.
-#endif freenas
-
 Windows automatically caches file sharing information. If changes are
 made to an SMB share or to the permissions of a volume/dataset being
 shared by SMB and the share becomes inaccessible, try logging out and
@@ -1636,7 +1627,9 @@ unless there is a specific need.**
   back up the data, create a new case-insensitive dataset, create an
   SMB share on it, then copy the data from the old one onto it. After
   the data has been checked and verified on the new share, the old one
-  can be deleted.
+  can be deleted. Once the case-insensitive dataset is created, the
+  share level auxiliary parameter “case sensitive = true” can be safely
+  set.
 
 * If present, remove options for extended attributes and DOS
   attributes in the share's
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1415,9 +1415,9 @@ This configuration screen is really a front-end to
    |                                  |                | :guilabel:`Hosts Deny` fields of a SMB share; uncheck if IP addresses are used to avoid the           |
    |                                  |                | delay of a host lookup                                                                                |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
-   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will support where the default sets automatic                 |
-   |                                  |                | negotiation; refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                      |
-   |                                  |                |                                                                                                       |
+   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will allow the client to use. Default selects "LANMAN1". SMB  |
+   |                                  |                | clients will automatically negotiate the highest supported protocol version that meets or exceeds     |
+   |                                  |                | this value ; refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions.                     |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
    | Server maximum protocol          | drop-down menu | the maximum protocol version the server will support; refer to                                        |
    |                                  |                | :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                                            |
@@ -1484,7 +1484,7 @@ This configuration screen is really a front-end to
    | SMB2_10        | used by Windows 7                                          |
    |                |                                                            |
    +----------------+------------------------------------------------------------+
-   | SMB3           | used by Windows 8                                          |
+   | SMB3           | Same as SMB3_11                                            |
    |                |                                                            |
    +----------------+------------------------------------------------------------+
    | SMB3_00        | used by Windows 8                                          |
@@ -1498,7 +1498,8 @@ This configuration screen is really a front-end to
    +----------------+------------------------------------------------------------+
 
 
-Changes to SMB settings and SMB shares take effect immediately.
+Changes to SMB settings take effect immediately. Changes to share settings will not
+effect until the client and server negotiate a new session. 
 
 .. note:: Do not set the *directory name cache size* as an
    :guilabel:`Auxiliary parameter`. Due to differences in how Linux
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1415,9 +1415,9 @@ This configuration screen is really a front-end to
    |                                  |                | :guilabel:`Hosts Deny` fields of a SMB share; uncheck if IP addresses are used to avoid the           |
    |                                  |                | delay of a host lookup                                                                                |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
-   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will allow the client to use; Default selects *LANMAN1*. SMB  |
-   |                                  |                | clients automatically negotiate the highest supported protocol version that meets or exceeds this     |
-   |                                  |                | value; refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                            |
+   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will support; default selects *LANMAN1*; SMB clients          |
+   |                                  |                | automatically negotiate the highest supported protocol version that meets or exceeds this value;      |
+   |                                  |                | refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                                   |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
    | Server maximum protocol          | drop-down menu | the maximum protocol version the server will support; refer to                                        |
    |                                  |                | :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                                            |
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1640,8 +1640,8 @@ unless there is a specific need.**
   overhead.
 
 
-  
-  
+
+
 .. index:: SNMP, Simple Network Management Protocol
 .. _SNMP:
 
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1415,9 +1415,9 @@ This configuration screen is really a front-end to
    |                                  |                | :guilabel:`Hosts Deny` fields of a SMB share; uncheck if IP addresses are used to avoid the           |
    |                                  |                | delay of a host lookup                                                                                |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
-   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will allow the client to use. Default selects "LANMAN1". SMB  |
-   |                                  |                | clients will automatically negotiate the highest supported protocol version that meets or exceeds     |
-   |                                  |                | this value ; refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions.                     |
+   | Server minimum protocol          | drop-down menu | the minimum protocol version the server will allow the client to use; Default selects *LANMAN1*. SMB  |
+   |                                  |                | clients automatically negotiate the highest supported protocol version that meets or exceeds this     |
+   |                                  |                | value; refer to :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                            |
    +----------------------------------+----------------+-------------------------------------------------------------------------------------------------------+
    | Server maximum protocol          | drop-down menu | the maximum protocol version the server will support; refer to                                        |
    |                                  |                | :numref:`Table %s <smb_protocol_ver_tab>` for descriptions                                            |
@@ -1484,7 +1484,7 @@ This configuration screen is really a front-end to
    | SMB2_10        | used by Windows 7                                          |
    |                |                                                            |
    +----------------+------------------------------------------------------------+
-   | SMB3           | Same as SMB3_11                                            |
+   | SMB3           | used by Windows 10; same as SMB3_11                        |
    |                |                                                            |
    +----------------+------------------------------------------------------------+
    | SMB3_00        | used by Windows 8                                          |
@@ -1498,8 +1498,8 @@ This configuration screen is really a front-end to
    +----------------+------------------------------------------------------------+
 
 
-Changes to SMB settings take effect immediately. Changes to share settings will not
-effect until the client and server negotiate a new session. 
+Changes to SMB settings take effect immediately. Changes to share settings only take
+effect after the client and server negotiate a new session. 
 
 .. note:: Do not set the *directory name cache size* as an
    :guilabel:`Auxiliary parameter`. Due to differences in how Linux
@@ -1537,6 +1537,7 @@ performance in some configurations and can be added to the
 a multiple of _SC_PAGESIZE (typically *4096*) to avoid memory
 fragmentation. This will increase Samba's memory requirements and
 should not be used on systems with limited RAM.
+#endif freenas
 
 Windows automatically caches file sharing information. If changes are
 made to an SMB share or to the permissions of a volume/dataset being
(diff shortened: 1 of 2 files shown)


anodos325:
Additional information about commit ab7d809 (Clarifications and corrections regarding SMB):

1.  When SMB config is changed through Services->SMB, the samba server is restarted.
2.  When share config is changed through Sharing->Windows (SMB) Shares-><share name>, the smb4.conf file is reloaded. 

(2) does not affect existing sessions while (1) does affect them. It's important to know what to expect on the back-end when making UI changes. ;-)

userguide/services.rst
```diff
@@ -1635,7 +1627,9 @@ unless there is a specific need.**
   back up the data, create a new case-insensitive dataset, create an
   SMB share on it, then copy the data from the old one onto it. After
   the data has been checked and verified on the new share, the old one
-  can be deleted.
+  can be deleted. Once the case-insensitive dataset is created, the
```
wblock: This seems to be in the wrong place.  Setting the parameter should be done right after creating the new share, before copying data.

Also, the "Once" construct can be confusing to non-native English speakers.  "When" or "After" is somewhat clearer.

userguide/sharing.rst
```diff
@@ -1057,9 +1057,9 @@ provides more details for each configurable option.
...
    |                              |               |          |                                                                                                             |
    +------------------------------+---------------+----------+-------------------------------------------------------------------------------------------------------------+
-   | Allow Guest Access           | checkbox      |          | if checked, no password is required to connect to the share and all users share the permissions of the      |
-   |                              |               |          | guest user defined in the :ref:`SMB` service                                                                |
-   |                              |               |          |                                                                                                             |
+   | Allow Guest Access           | checkbox      |          | if checked, no password is required to connect to the share. User login attempts with a bad password are    |
```
wblock: See note above about sentences in tables.  "No password is required" is less clear than "a password is not required".  "User login attempts" is not clear. A user does not log in to a share, they connect to it.

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
-  checked (the default), other users will see the names of every share
-  that exists using Windows Explorer, but they will receive a
-  permissions denied error message if they try to access someone
-  else's share. If this box is unchecked, even the owner of the share
-  will not see it or be able to create a drive mapping for the share
-  in Windows Explorer. However, they can still access the share from
-  the command line. Unchecking this option provides limited security
-  and is not a substitute for proper permissions and password control.
+* When the :guilabel:`Browsable to Network Clients` box is checked (the 
+  default), the share will be visible in the server's browse list. If the share
```
wblock: "visible in the server's browse list" is not clear.  Also, use "is" instead of "will be".

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
-  that exists using Windows Explorer, but they will receive a
-  permissions denied error message if they try to access someone
-  else's share. If this box is unchecked, even the owner of the share
-  will not see it or be able to create a drive mapping for the share
-  in Windows Explorer. However, they can still access the share from
-  the command line. Unchecking this option provides limited security
-  and is not a substitute for proper permissions and password control.
+* When the :guilabel:`Browsable to Network Clients` box is checked (the 
+  default), the share will be visible in the server's browse list. If the share 
+  is is flagged as a :guilabel:`home share`, unchecking this box will hide the
```
wblock: Duplicate "is" at start of line.
Also, s/will hide/hides/

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
-  permissions denied error message if they try to access someone
-  else's share. If this box is unchecked, even the owner of the share
-  will not see it or be able to create a drive mapping for the share
-  in Windows Explorer. However, they can still access the share from
-  the command line. Unchecking this option provides limited security
-  and is not a substitute for proper permissions and password control.
+* When the :guilabel:`Browsable to Network Clients` box is checked (the 
+  default), the share will be visible in the server's browse list. If the share 
+  is is flagged as a :guilabel:`home share`, unchecking this box will hide the 
+  [homes] share so that only the dynamically generated home share with the
```
wblock: Markup seems needed for [homes], but it's not clear to me what is being denoted there.  Literal value, variable value, setting in configuration file...

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
+  authenticated user's name will be visible (in contrast with [homes]
```
wblock: In general, avoid possessives.  It is usually not needed: "username" rather than "user's name", for example.

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
+  and [<username>] both being visible).
```
wblock: Again, markup is needed.

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
-  the command line. Unchecking this option provides limited security
-  and is not a substitute for proper permissions and password control.
+* When the :guilabel:`Browsable to Network Clients` box is checked (the 
+  default), the share will be visible in the server's browse list. If the share 
+  is is flagged as a :guilabel:`home share`, unchecking this box will hide the 
+  [homes] share so that only the dynamically generated home share with the
+  authenticated user's name will be visible (in contrast with [homes]
+  and [<username>] both being visible). 
+
+  The presence of a share in a user's browse list does not grant the user
```
wblock: Hard to tell what you mean by "a user's browse list".  Does this mean "Shares that appear in the File Explorer" (or whatever Windows calls it today) "do not necessarily mean the user has read or write permissions to them."?

userguide/sharing.rst
```diff
@@ -1092,16 +1092,17 @@ settings:
...
+  default), the share will be visible in the server's browse list. If the share 
+  is is flagged as a :guilabel:`home share`, unchecking this box will hide the 
+  [homes] share so that only the dynamically generated home share with the
+  authenticated user's name will be visible (in contrast with [homes]
+  and [<username>] both being visible). 
+
+  The presence of a share in a user's browse list does not grant the user  
+  read or write permissions on the share. Users are still able to directly 
+  access and map shares that are not visible in the browse list by using 
+  the share's UNC path. Hence, this option is of limited security utility.
```
wblock: "This option provides no real security because shares that are not visible in the File Explorer (or whatever) can still be accessed with a *UNC* path."

Not sure about markup, but I think that's what this is trying to say.

userguide/sharing.rst
```diff
@@ -1133,13 +1134,7 @@ like Windows 7 will not be able to connect with NTLMv1 disabled.
...
 <https://support.microsoft.com/en-us/help/2793313/security-guidance-for-ntlmv1-and-lm-network-authentication>`_
 has information about the security implications and ways to enable
-NTLMv2. If changing the client configuration is not possible, NTLMv1
-authentication can be enabled by adding this entry to
-:guilabel:`Auxiliary Parameters`:
-
-.. code-block:: none
-
-   ntlm auth = yes
+NTLMv2.
```
wblock: This removes the mention of NTLMv1 entirely.  Shouldn't it point to the checkmark option?

userguide/sharing.rst
```diff
@@ -1292,27 +1287,27 @@ for more details.
    | xattr_tdb           | stores Extended Attributes (EAs) in a tdb file so they can be used on filesystems which do not provide support for EAs                     |
    |                     |                                                                                                                                            |
    +---------------------+--------------------------------------------------------------------------------------------------------------------------------------------+
-
+   | zfs_space           | correctly calculates ZFS space used by share, including space used by ZFS snapshots, quotas, and resevations. Enabled by default.          |
```
wblock: No full sentences in table descriptions.

userguide/sharing.rst
```diff
@@ -1292,27 +1287,27 @@ for more details.
...
+   |                     |                                                                                                                                            |
+   +---------------------+--------------------------------------------------------------------------------------------------------------------------------------------+
+   | zfsacl              | provide ACL extensions for proper integration with ZFS. Enabled by default.                                                                |
```
wblock: As above, no full sentences.  (I don't know why, it was that way when I got here.)

userguide/sharing.rst
```diff
@@ -1292,27 +1287,27 @@ for more details.
...
 
 .. note:: When using :guilabel:`fruit`, also add the
    :guilabel:`streams_xattr` and :guilabel:`catia` VFS objects and be
    sure to configure **all** SMB shares this way. Reboot the Mac
    client after making this change.
 
-These VFS objects do not appear in the drop-down menu as they are
-always enabled:
+These VFS objects do not appear in the drop-down menu and are dynamically
+configured:
```
wblock: What does "dynamically configured" mean?  Set by the system, set by the user at runtime?

userguide/sharing.rst
```diff
@@ -1292,27 +1287,27 @@ for more details.
...
 
 * **recycle:** moves deleted files to the recycle directory instead of
-  deleting them
+  deleting them. Controlled by a checkbox in the share configuration.
```
wblock: Second sentence needs to use :guibutton: or a :ref: to the appropriate section.

userguide/services.rst
```diff
@@ -1484,7 +1484,7 @@ This configuration screen is really a front-end to
    | SMB2_10        | used by Windows 7                                          |
    |                |                                                            |
    +----------------+------------------------------------------------------------+
-   | SMB3           | used by Windows 8                                          |
+   | SMB3           | Same as SMB3_11                                            |
```
anodos325: I see your point. Since SMB3 = SMB3_11, "used by Windows 8" is technically incorrect. This is relevant when setting server minimum protocol = SMB3.  
Does the following look better?
**used by Windows 10; same as SMB3_11**

userguide/services.rst
```diff
@@ -1537,15 +1538,6 @@ a multiple of _SC_PAGESIZE (typically *4096*) to avoid memory
...
 
-To increase network performance, read the Samba section on socket
-options in the
-`smb.conf manual page
-<https://www.freebsd.org/cgi/man.cgi?query=smb.conf&manpath=FreeBSD+11.0-RELEASE+and+Ports>`_.
-It indicates which options are available and recommends that you
-experiment to see which are supported by your clients and improve your
-network's performance.
-#endif freenas
-
```
anodos325: Heh. I'm kind of new to this git/github thing and didn't realize any additional edits I make get appended to existing pull requests. :-)

Adjusting socket options is somewhat frowned upon in the samba mailing lists (JRA has even mentioned ripping the parameter out of samba entirely). Most edits people do are voodoo, and some of them actually hurt performance. It's probably better to adjust send and receive buffers through sysctls (which people with Intel 10GB NICs tend to do). This is because it will still allow the kernel to automatically tune send and receive buffers. 

Manipulating IPTOS headers may help some people who have samba passing through routers (and have configured QOS in routers accordingly), but I believe this probably is a more advanced topic than most people looking at the FreeNAS handbook are trying to address.

TL;DR, socket options aren't exactly where I'd point people to if they were trying to improve samba performance.

I'll fix the #endif freenas. I'm still learning the ropes of the markup language. I'll also leave the blurb about socket options if you want ;-)

userguide/services.rst
anodos325: My writing can also confuse native English speakers. :) The problem with the way that it is written originally is that no performance improvement will occur until the admin disables the case sensitivity checks. 

Another way of looking at it is that case sensitivity checks cause o(n^2) delay. Disabling them speeds things up. Making the dataset case insensitive allows you to do this safely. 

I'm open to better ways to say this.

userguide/sharing.rst
anodos325: Sorry. This was half-baked. [homes] is a special section of the smb.conf file that is generated when when the user checks the box "Use as home share". Enabling home shares in samba in FreeNAS will basically generate two shares:
\\freenas\homes
\\freenas\%U
Unchecking "browsable" for a home share makes it so that \\freenas\homes isn't visible, but \\freenas\%U will be visible.

userguide/sharing.rst
anodos325: In my defense, it made sense when I was writing it at 4:30 AM before drinking coffee.

userguide/sharing.rst
b-a-t: If you dive into the protocol version differences, it's worth mentioning, that most of the time when you talk about SMB2, you should mean SMB2_02. See [#24334](https://bugs.freenas.org/issues/24334).

>     SMB1 -> NT1
>     SMB2 -> SMB2_02
>     SMB2.1 -> SMB2_10
>     SMB3 -> SMB3_00

I hope to fix that in a GUI one day, around 11.1, I hope, but meanwhile clarifying it in the docs could be an option.

anodos325: I believe that **server min protocol = SMB2** is identical to **server min protocol  = SMB2_10**
This means that Vista clients will fail to negotiate a supported protocol. Yay for inconsistency! 

Perhaps we need a errata section (or separate book) for weird SMB client behavior.

userguide/sharing.rst
```diff
@@ -1093,16 +1093,15 @@ settings:
...
-  [homes] share so that only the dynamically generated home share with the
-  authenticated user's name will be visible (in contrast with [homes]
-  and [<username>] both being visible). 
-
-  The presence of a share in a user's browse list does not grant the user  
-  read or write permissions on the share. Users are still able to directly 
-  access and map shares that are not visible in the browse list by using 
-  the share's UNC path. Hence, this option is of limited security utility.
+  default), the share is visible when the user lists the available shares 
+  through Windows File Explorer or through the *net view* command. If the
```
wblock: The markup for a command is :command:`net view`.

userguide/sharing.rst
```diff
@@ -1093,16 +1093,15 @@ settings:
...
-  authenticated user's name will be visible (in contrast with [homes]
-  and [<username>] both being visible). 
-
-  The presence of a share in a user's browse list does not grant the user  
-  read or write permissions on the share. Users are still able to directly 
-  access and map shares that are not visible in the browse list by using 
-  the share's UNC path. Hence, this option is of limited security utility.
+  default), the share is visible when the user lists the available shares 
+  through Windows File Explorer or through the *net view* command. If the  
+  share is flagged as a :guilabel:`home share`, unchecking this box hides the
```
wblock: guilabel is for a GUI element like a button or list.  I'm not sure what "home share" represents here.  Well, a concept, but not a literal button.  I think.

userguide/sharing.rst
```diff
@@ -1093,16 +1093,15 @@ settings:
...
-  The presence of a share in a user's browse list does not grant the user  
-  read or write permissions on the share. Users are still able to directly 
-  access and map shares that are not visible in the browse list by using 
-  the share's UNC path. Hence, this option is of limited security utility.
+  default), the share is visible when the user lists the available shares 
+  through Windows File Explorer or through the *net view* command. If the  
+  share is flagged as a :guilabel:`home share`, unchecking this box hides the 
+  *homes* share so that only the dynamically generated home share with the
+  authenticated username will be visible (in contrast with a share named
+  *homes* and the dynamically-generated home share both being visible).
```
wblock: Sorry, I can't follow this explanation.  It's confusing have a share named "homes" and using the term "home share".

userguide/sharing.rst
```diff
@@ -1287,10 +1288,10 @@ for more details.
    | xattr_tdb           | stores Extended Attributes (EAs) in a tdb file so they can be used on filesystems which do not provide support for EAs                     |
    |                     |                                                                                                                                            |
    +---------------------+--------------------------------------------------------------------------------------------------------------------------------------------+
-   | zfs_space           | correctly calculates ZFS space used by share, including space used by ZFS snapshots, quotas, and resevations. Enabled by default.          |
+   | zfs_space           | correctly calculates ZFS space used by share, including space used by ZFS snapshots, quotas, and resevations; enabled by default           |
```
wblock: I think this would read better with "the share" instead of just "share".

userguide/sharing.rst
```diff
@@ -1299,14 +1300,15 @@ for more details.
    sure to configure **all** SMB shares this way. Reboot the Mac
    client after making this change.
 
-These VFS objects do not appear in the drop-down menu and are dynamically
-configured:
+These VFS objects do not appear in the selection box:
 
 * **recycle:** moves deleted files to the recycle directory instead of
-  deleting them. Controlled by a checkbox in the share configuration.
+  deleting them. Controlled by :guilabel:`Export Recycle Bin`.
```
wblock: Good!  It should still point the reader to where that checkbox is, so:

Controlled by :guilabel:`Export Recycle Bin` in the :ref:`SMB share configuration <smb_share_opts_tab>`.

userguide/sharing.rst
```diff
@@ -1299,14 +1300,15 @@ for more details.
...
 
 * **shadow_copy2:** a more recent implementation of
-  :guilabel:`shadow_copy` with some additional features
+  :guilabel:`shadow_copy` with some additional features. *shadow_copy2*
+  and associated *shadow:* parameters are automatically added to the
```
wblock: Why the colon in *shadow:*?

userguide/sharing.rst
anodos325: Well, the behavior is confusing as well. ;-) For instance, in my smb.conf I have the share:
[my share]
path = /mnt/dozer/share

It appears in explorer as "My Share". If I check the box "use as a homeshare", the smb4.conf is rewritten so that it has 
[homes]
path = /mnt/dozer/share

After checking the button I have two shares in windows explorer "homes" and "awalker" (my username). If I uncheck "browsable to network clients" in the share configuration, I will only have one share "awalker" because the automatically generated home share inherits the browsable parameter from the [global] section of the smb4.conf.

So the behavior is somewhat complicated and confusing for users who aren't familiar with Samba. [homes] is a special reserved section of the smb.conf file that has this unique behavior. 

So "homes" is a share. "Use as a home share" is a gui element checkbox. "Home share" refers to the share that has been flagged in the FreeNAS UI as a home share.

userguide/sharing.rst
anodos325: Some VFS modules in samba take additional parameters as defined in the module. FreeNAS sets the following shadow_copy2 parameters:

- shadow:snapdirseverywhere = yes
- shadow:format = <snapshot task>
- shadow:localtime = yes
- shadow:sort = desc
- shadow:snapdir = .zfs/snapshot

I tried to just write "shadow:*" but wasn't sure how the markup language would like *shadow:**

anodos325: I decided to simplify this and remove reference to "shadow:*". I.e. "and associated parameters"

userguide/sharing.rst
anodos325: I was probably commenting in the Russian forums when I wrote that. :-) English and its silly articles.

userguide/sharing.rst
anodos325: Tried to clarify. Perhaps it's better to just create a separate explanation for "use as a home directory"

Issue #18, opened 2017-05-21
Link to solarisinternals "ZFS Storage Pools Recommendations" is dead


alanwsmith:
[Section 1.3.3 (Storage Disks and Controllers)](https://doc.freenas.org/9.10/intro.html#storage-disks-and-controllers) of the current 9.10 documentation contains:

> Users new to ZFS who are purchasing hardware should read through ZFS Storage Pools Recommendations first.

with a link to:

`http://www.solarisinternals.com/wiki/index.php/ZFS_Best_Practices_Guide#ZFS_Storage_Pools_Recommendations`

That link appears dead. (I believe the entire `solarisinternals.com` site is now defunct.)

The Wayback Machine has an archive:

`https://web.archive.org/web/20161028084224/http://www.solarisinternals.com/wiki/index.php/ZFS_Best_Practices_Guide#Storage_Pools`

While using that link might not be the best approach, it's at least an option.

PR #22, opened 2017-07-15
Fixes GiHub Issue #18 and adds information regarding FreeNAS Bug #23977

Candunc:
https://github.com/freenas/freenas-docs/issues/18
https://bugs.freenas.org/issues/23977


wblock:
For the future, instructions are generally improved by avoiding the informal "you".  So rather than "If you are using the RealVNC Client ... you must disable the...", say something like "When using the RealVNC Client ... disable the..."

Thanks!

PR #29, opened 2017-10-16
Added: Requirements for the HA x509 certs

650elx:
Please reformat this as needed.


wblock:
Added separately.  Thanks!

PR #34, opened 2017-11-13
Begin reviewing vCenter documentation:

Mrt134:
See commit message for full details.


userguide/tn_vcenter.rst
```diff
@@ -132,7 +132,8 @@ in the left tree. In the secure connection screen shown in
 Using the vCenter Plugin
 ------------------------
 
-From the main vSphere window, click
+Log in to the paired vCenter Host by typing its IP address into a
```
wblock: Do not use possessives unless there is no other alternative.  And there is always an alternative.

userguide/tn_vcenter.rst
```diff
@@ -228,9 +229,11 @@ shows the configuration screen for the new datastore:
    Configuring a VMFS Datastore
 
 
-Enter a name for the new datastore and the desired size. Enter the IP
-address of the %brand% system in the :guilabel:`Data Path IP` box.
-Select the desired :guilabel:`VMFS Version` from the drop-down. If a
+Enter a name for the new datastore and the desired size. Datastore names
+can only contain alphanumeric characters and only the :literal:`-` and
```
wblock: This is somewhat oddly worded.  First it says "can only contain alphumeric" and then "and only the - and ."

userguide/tn_vcenter.rst
```diff
@@ -42,7 +42,7 @@ shown in
 
 .. _vcenter_plugin_config_fig:
 
-.. figure:: images/tn_vcenter-plugin-config.png
+.. figure:: images/tn_vcenter-plugin-config_a.png
```
wblock: Let's not rename this image, just replace it with the same name.  Yes, we need to update the process in ix-docs.

PR #37, opened 2017-11-19
Storage isn't very clear about mirrored device management, add some short fixes on key points.

stilez:
For example, when the docs say something "isn't possible" (like adding a disk), it may be possible for a mirror or single HDD but not for RAIDZ, and that can confuse users. Users with mirrors might try to find about managing them, which isn't covered, or think that "extending" covers "adding mirrors" which it doesn't.

I've tried to briefly improve this area. Please edit if it can be done better

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -65,7 +65,14 @@ Volume Manager
 old data on added disks is overwritten, so save it elsewhere before
 reusing a disk. Please see the :ref:`ZFS Primer` for information on
 ZFS redundancy with multiple disks before using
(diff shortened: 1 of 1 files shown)


wblock:
Thank you for submitting this.  The present problem is that mirrors created through the GUI might not be consistent.  However, there is a need to add this to the GUI, and I have created https://redmine.ixsystems.com/issues/26760 to track that.

Also, this made me review and rewrite a fair portion of the storage chapter, which it badly needed and improved it greatly.  Thanks!

PR #44, opened 2018-01-11
Various edits to the "encryption" header text

stilez:
1) Link to a more exact description about GELI key management.

People using encryption (and managing its various keys) are likely to be the kind of people who will want exact details how keys are stored and managed. While they could Google GELI, that won't explain how %brand% uses and stores them, and is a lot to read through. However there's a nice and concise forum post which summarises it nicely and worth linking to for that info.....

2) Move "encryption performance" to its own small heading below the bullet list

Performance is always a question/concern with encryption. This keeps the main bullet list short and "to the point", and pulls the 1 or 2 bullets on performance to below the list, which (to me) flows much better.

3) Merge "encrypted on write" and "decrypted on fly during read" to a previous bullet where they fit nicely

4) Remove an "#ifdef freenas" - this seems to apply universally not just to specific %brand%s as it's generic to any hw and GELI use. If incorrect please edit to fix!

5) Cleaner separation between CPU AES-NI impact and disk count impact, and reduce redundant wording (we mention "multiple disk degradation" twice and blur it with AES-NI impact)

6) Mention whether LOG (ZIL) is encrypted or not, since we already mention everything else (RAM, L2ARC and SWAP).

7) The section on "managing encryption" describes the existence and controls over keys and passphrases, but doesn't tell the user what each is, and what they do. Fairly crucial information to be clear about before we tell the user about buttons that are shown to manage them :)

8) Mention (in a note) that it's rarely useful to back up a disk's master keys - because this is a very obvious and usual concern/question with full disk encryption systems, and the answer isn't very obvious. No need for users to be worried if considering use of encryption.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1290,11 +1290,10 @@ Managing Encrypted Volumes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 %brand% automatically generates a randomized **encryption key** 
-whenever a new encrypted volume is created. The data "at rest" (on disk)
-on an encrypted volume is **always encrypted**, but access to the decrypted
-data on a live (running) system can be customized for each volume.
+whenever a new encrypted volume is created. %brand% needs this key 
+to read and decrypt any data (or any other information) within the volume.
 
-By default, encryption keys will be stored locally within %brand*'s data 
+By default, encryption keys will be stored locally within %brand%'s data 
 files. They can also be downloaded as a safety measure, to allow
 decryption on a different system in the event of failure, or to allow
 the locally stored key to be deleted for extra security. Encryption keys
@@ -1311,12 +1310,15 @@ implications are:
   data. If a passphrase is set, this must **also** be provided before data 
   can be accessed (`two factor authentication <https://en.wikipedia.org/wiki/Multi-factor_authentication>`_).
 
-Decrypted data **cannot be accessed** (other than from *L2ARC*) when the
-disks are removed, the system is shut down, or (on a running system) when 
-the key is unavailable. If the key is protected with a passphrase, then 
-data cannot be decrypted without having both key and passphrase. 
-Decryption is per-volume not per-user, so when a volume is unlocked, data 
-will be decrypted for *any* user whose permissions allow them to access it.
+Data stored on an encrypted volume disk is **always encrypted**. Data to be 
+written to an encrypted volume is encrypted when written to the LOG disk (ZIL),
+but is **not** stored encrypted in L2ARC, if present. With the exception of L2ARC, 
+decrypted data **cannot be accessed** when the disks are removed, the system
+has been shut down, or (on a running system) when the volume is 'locked' and
+the key is unavailable. If the key is protected with a passphrase, then data
+cannot be decrypted without having both key and passphrase. Decryption is 
+per-volume not per-user, so when a volume is unlocked, data will be decrypted
+for *any* user whose permissions allow them to access it.
 
 .. note:: By design, `GELI <http://www.freebsd.org/cgi/man.cgi?query=geli>`_
    uses *two* randomized encryption keys for each disk. One is the key discussed
@@ -1325,13 +1327,13 @@ will be decrypted for *any* user whose permissions allow them to access it.
    a disk's master key due to disk corruption would be equivalent to any other
    disk failure, and in a redundant pool, other disks will contain accessible 
    copies of the uncorrupted data.
-   Therefore, while it is *possible* to separately backup any master keys, 
-   it is **not** usually considered necessary or useful to do so.
+   Therefore, while it is *possible* to separately back up any master keys, 
+   it is not usually considered necessary or useful to do so.
 
 
-.. _Additional controls for encrypted volumes:
+.. _Additional Controls for Encrypted Volumes:
 
-Additional controls for encrypted volumes
+Additional Controls for Encrypted Volumes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 If the :guilabel:`Encryption` box is checked during the creation of a
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -212,9 +212,9 @@ the details when considering whether encryption is right for your
   on top of the encrypted devices. As data is written, it is automatically 
   encrypted, and as data is read, it is decrypted on the fly. 
 
-* Data in the ARC cache and the contents of RAM are **always unencrypted**, 
-  and Swap is **always encrypted** (even on unencrypted volumes). The 
-  LOG (ZIL) cache is **encrypted if it relates to an encrypted disk**.
+* Data in memory, including ARC, is **not** encrypted. ZFS data on disk,
+  including ZIL and SLOG, are encrypted if the underlying disks **are** 
+  encrypted. Swap data on disk is **always** encrypted.
 
 * This type of encryption is primarily targeted at users who store
   sensitive data and want to retain the ability to remove disks from
@@ -230,10 +230,8 @@ the details when considering whether encryption is right for your
   inaccessible. Always back up the key!
 
 * The encryption key is per ZFS volume (pool). Multiple pools each
-  have their own encryption key. 
-  
-* Technical details about how encryption keys are used, stored and managed
-  within %brand% can be found in
+  have their own encryption key. Technical details about how encryption
+  keys are used, stored and managed within %brand% can be found in
   `this forum post <https://forums.freenas.org/index.php?threads/recover-encryption-key.16593/#post-85497>`_.
 
 * There is no way to convert an existing, unencrypted volume. Instead,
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -225,27 +230,11 @@ the details when considering whether encryption is right for your
   inaccessible. Always back up the key!
 
 * The encryption key is per ZFS volume (pool). Multiple pools each
-  have their own encryption key.
-
-#ifdef freenas
```
wblock: This #ifdef is needed.  All TrueNAS hardware has AESNI, and does not need this section.

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
 Managing Encrypted Volumes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+%brand% automatically generates a randomized **encryption key** 
+whenever a new encrypted volume is created. The data "at rest" (on disk)
+on an encrypted volume is **always encrypted**, but access to the decrypted
+data on a live (running) system can be customized for each volume.
```
wblock: This is confusing.  It seems to be an introduction to when the encryption keys can be entered, but the word "access" can be read to mean several other things.

userguide/storage.rst
```diff
@@ -279,6 +265,29 @@ A pop-up message shows a reminder that
...
+Encryption performance
+^^^^^^^^^^^^^^^^^^^^^^
+
+If the processor supports the 
+`AES-NI <https://en.wikipedia.org/wiki/AES-NI#Supporting_CPUs>`_
+instruction set, there is very little, if any, degradation in
+performance when using encryption and only a few disks.
+Performance will suffer if the CPU does not support
+AES-NI or if no crypto hardware is installed.  Without hardware
+acceleration, there will be about a 20% performance decrease for a
```
wblock: I'm hesitant to give a flat 20% number here, since it is likely to be dependent on more than one factor.

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
 Managing Encrypted Volumes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+%brand% automatically generates a randomized **encryption key** 
+whenever a new encrypted volume is created. The data "at rest" (on disk)
+on an encrypted volume is **always encrypted**, but access to the decrypted
+data on a live (running) system can be customized for each volume.
+
+By default, encryption keys will be stored locally within %brand*'s data
```
wblock: Typo, %brand*

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
...
+  and always accessible when system running. (Protects "data at rest" only.)
+
+* *Key stored locally, with passphrase* - user must provide passphrase
+  before anyone can access data.
+
+* *Key not stored locally* - user must provide key before anyone can access
+  data. If a passphrase is set, this must **also** be provided before data 
+  can be accessed (`two factor authentication <https://en.wikipedia.org/wiki/Multi-factor_authentication>`_).
+
+Decrypted data **cannot be accessed** (other than from *L2ARC*) when the
```
wblock: The reader will ask "how can I access my lost data in the L2ARC?", so I would remove that aside.  Incidentally, asides should be avoided as an interruption in the sentence.

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
...
+will be decrypted for *any* user whose permissions allow them to access it.
+
+.. note:: By design, `GELI <http://www.freebsd.org/cgi/man.cgi?query=geli>`_
+   uses *two* randomized encryption keys for each disk. One is the key discussed
+   in this guide. The other, called the disk's "master key", is stored on the
+   disk itself, in a strongly encrypted form which the user never sees. Loss of 
+   a disk's master key due to disk corruption would be equivalent to any other
+   disk failure, and in a redundant pool, other disks will contain accessible 
+   copies of the uncorrupted data.
+   Therefore, while it is *possible* to separately backup any master keys,
```
wblock: s/backup/back up/

"backup" is a noun, "back up" is the verb.  I can prove this: nobody says "I am backuping this data".

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
...
+
+.. note:: By design, `GELI <http://www.freebsd.org/cgi/man.cgi?query=geli>`_
+   uses *two* randomized encryption keys for each disk. One is the key discussed
+   in this guide. The other, called the disk's "master key", is stored on the
+   disk itself, in a strongly encrypted form which the user never sees. Loss of 
+   a disk's master key due to disk corruption would be equivalent to any other
+   disk failure, and in a redundant pool, other disks will contain accessible 
+   copies of the uncorrupted data.
+   Therefore, while it is *possible* to separately backup any master keys, 
+   it is **not** usually considered necessary or useful to do so.
```
wblock: Emphasis not really necessary.

userguide/storage.rst
```diff
@@ -1280,6 +1289,51 @@ changed, and destroying a zvol requires confirmation.
...
+
+
+.. _Additional controls for encrypted volumes:
+
+Additional controls for encrypted volumes
```
wblock: This should be in title caps.

userguide/storage.rst
stilez: If there's a page you can link which gives the preprocessor directives usable in this doc, it would help - I can see there's an `#ifdef ... #endif`, but is the matching syntax `#ifdef ... #else ... #endif`, or `#ifndef ... #endif` for any alternate wording?

userguide/storage.rst
stilez: It was in the original text so I left it. Let me know any preferred wording.  But overall my gut feeling is that it's better to give some "ballpark" indication (1%? 5%? 70%?) than none at all, as many just want "some idea" whether it's a showstopper or not.   Users who need more certainty will probably benchmark.

userguide/storage.rst
stilez: Will fix by merging into para below bullets, which already mentions that topic so it also avoids possible duplication.

userguide/storage.rst
stilez: ok

userguide/storage.rst
stilez: The problem is that, if that isn't made very clear indeed, then a user will get the message "decrypted data cannot be accessed", and probably not link it to the mention some distance up, that L2ARC isn't ever encrypted. Especially given the price of large SSD's, a user who buys any L2ARC is likely to have quite a lot of it (and 250GB will be quite a lot from a "leakable data" perspective). So the reality is that a lot of the pool's most commonly/recently accessed files will present _no encryption barrier at all_. The docs need to be very clear/explicit on that point in the section for users considering encrypted volumes, in the context of any statement suggesting that data at rest is secure. 

If the concern is _"user with data loss may wonder if lost data can be recovered from their L2ARC"_, then maybe there's a better way to say "not really" in the "lost data" section of the guide?

userguide/storage.rst
stilez: No proof needed. Will fix.

userguide/storage.rst
stilez: ok

userguide/storage.rst
stilez: I'll have to look that up. If it's obvious then ok, if not, syntax?

stilez:
@wblock - Done what I can until I hear back from you. to finish the changes I need pre-processor syntax, text you prefer for "20%" figure currently in guide, and what you reckon about L2ARC. The rest are done.

userguide/storage.rst
wblock: In the main directory, run `./preproc -h` for the full help information on the preprocessor.  The #ifdef and #endif terms match: in this case, it would be `#ifdef truenas` and `#endif truenas`.

userguide/storage.rst
```diff
@@ -209,7 +209,12 @@ the details when considering whether encryption is right for your
 
 * This is full disk encryption and **not** per-filesystem encryption.
   The underlying drives are first encrypted, then the pool is created
-  on top of the encrypted devices.
+  on top of the encrypted devices. As data is written, it is automatically 
+  encrypted, and as data is read, it is decrypted on the fly. 
+
+* Data in the ARC cache and the contents of RAM are **always unencrypted**, 
+  and Swap is **always encrypted** (even on unencrypted volumes). The 
+  LOG (ZIL) cache is **encrypted if it relates to an encrypted disk**.
```
wblock: After technical review, here is what we came up with:

*Data in memory, including ARC, is not encrypted. ZFS data on disk, including ZIL and SLOG, are encrypted if the underlying disks are encrypted. Swap data on disk is always encrypted.*

stilez:
Updated - anything more needed?

PR #45, opened 2018-01-17
This will add in rotating ad functionality

JoshDW19:
This will add in rotating ad functionality and will dynamically hide the ads if the TAG=truenas flag is used.


wblock:
Closing this old version.  The new one should not add separate CSS file for TrueNAS.

PR #46, opened 2018-01-24
Rsync service + tasks

stilez:
See https://redmine.ixsystems.com/issues/21374 . 

1) A lot of the detail under rsync tasks was really more about rsync background, and overlapped with the service page.  So I've collected all the rsync background, basics and terminology at "Rsync service" for user ease, which simplifies the "tasks" page considerably.

2) A lot of the text on Rsync wasn't helpful at all. (See redmine link above). I've tried to tune it a little without adding much or getting into detailed config, so a user gets a better idea of this valuable service and the terminology used by *NAS.

3) The guide doesn't cover two rather crucial "need to know"s about Rsync - It doesn't encrypt data, so use some kind of secure transport (SSH) across insecure networks or internet if this matters, and use a minimally privileged account for the task - and especially not your root user. I don't know to what extent middleware will check someone using an overprivileged account (even if not root), but both of these warnings should probably be explicitly stated, however briefly.

I'm not 100% sure the correct/preferred terminology so this is bound to need some tidying up, and perhaps some minor fixes if I've missed anything. Please be gentle! 

But overall, this PR should help users on this valuable service without overstraying into additional rsyncd.conf material.

### DIFF
--- a/userguide/services.rst
+++ b/userguide/services.rst
@@ -1061,12 +1061,13 @@ that has greater access or privileges than the transfer requires.
 Particularly, **do not use the root account for Rsync** without carefully
 considering security imlications. 
 
-Rsyncd also supports "forced commands" - configurations where a remote
+Rsyncd can also be configured to create "forced commands", in which a remote
 user can only control some pre-defined session parameters (or none of them), 
-all other session properties being defined within the rsyncd configuration
-of the %brand% server. These are outside the scope of this Guide but allow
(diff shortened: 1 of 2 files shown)


wblock:
For the future:

1. Please make sure not to leave whitespace at the end of lines.  Some editors can be configured to show this.

2. Use two underscores after external URL links to prevent Sphinx from trying to add them as local labels.

PR #57, opened 2018-03-09
Edit "dedup" section

stilez:
The general info repeats a lot (memory amounts, inability to undedup existing data, warnings), is inaccurate in part (dedup only needs 1 - 2 GB per TB **if used when actually appropriate**, i.e., when there are high duplicate data levels), and misses out crucial detail (how to tell how much dup data you have, and the fact that other ARC usage can push DDT metadata out of RAM!)

I've grouped "like with like" and organised it - hopefully better - and added the missing info.

I'm not sure how to format the warning paragraphs, as a lot of the "be careful and try to avoid dedup" is really warning text. Perhaps a separate section? Maybe you can suggest a better way. 

I also don't remember how to format the link to "Tunable/Tunables", can you advise or correct?

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -793,7 +793,7 @@ GB of RAM per TB of deduplicated storage. The exact size required for
 deduplication data is found by using the :command:`CLI` command 
 :command:`zdb -U /data/zfs/zpool.cache -D pool_name` to identify the 
 total number of blocks in the pool, and the number of bytes of RAM 
-required per block (shown as *"{number) in core"* in the output). The 
+required per block (shown as *"{number} in core"* in the output). The 
 command  :command:`zdb -U /data/zfs/zpool.cache -S pool_name` 
 displays an estimate of the storage saving if deduplication is 
 applied to a dataset. Note that these commands may take a long time 
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -723,60 +723,61 @@ Deduplication
 Deduplication ("dedup") is a process where ZFS aims to store a single copy
 of any data that is repeated in different blocks or files, rather than storing 
 identical copies each time the data occurs. Deduplication 
-potentially allows repeated and identical data to be stored very 
-efficiently in a smaller space, but also places a 
-very heavy burden on RAM and on the CPU. In particular, dedup 
-significantly reduces write performance and usually requires large (and at 
-times huge) amounts of RAM.  See :ref:`Deduplication RAM requirements` for details.
-Because of the demands that dedup places on a system and the impact on performance, 
-enabling dedup is **usually not recommended** and the setting should be left in
-the default setting ("off") in most situations.
+allows repeated and identical data to be stored very 
+efficiently in a smaller space. 
+
+Deduplication can be enabled during dataset creation, and 
+occurs at the time when files are read and written. The setting can 
+be changed at any time but will not affect data already written to 
+the pool.  
+
+Deduplication can require **very large amounts of RAM** and substantial impacts write 
+performance and CPU usage.  See :ref:`Deduplication RAM requirements` for details.
+
+Due to the demands that deduplication places on a system and the impact on performance,
+deduplication is **off by default**. Enabling deduplication without good reason is
+**not usually recommended** unless there are good reasons for doing so, the system has 
+sufficient RAM and CPU, and the performance impact has been tested on a sample of data and
+is acceptable.
 
 .. warning::
 
    The more data written to a deduplicated dataset, the more RAM it 
-   requires. The DDTs (dedup tables) also compete with other data for RAM,
-   and can be pushed out of RAM in some cases. When the system starts 
-   storing DDTs on disk because they no longer fit into RAM or have been
-   pushed out of RAM, performance craters.
+   requires. The deduplication tables (DDT) also compete with other data for RAM,
+   and can be pushed out of RAM and onto disk in some cases. Performance will be 
+   unacceptable if DDTs cannot be managed entirely in RAM.
    
-   Importing an unclean deduplicated pool can also require large amounts of RAM, 
-   so that the deduplicated data in the pool can be checked during import. 
-   If the system does not have the needed RAM, it will panic. The only 
-   solution then is to add more RAM or recreate the pool.
+Importing an unclean deduplicated pool can also require large amounts of RAM, 
+so that the deduplicated data in the pool can be checked during import. A system
+panic will occur if the import process runs out of RAM while running. The only 
+solution then is to add more RAM or recreate the pool.
 
-   **There is also no quick way to undedup data within a dataset once deduplication 
-   is enabled, since disabling deduplication DOES NOT CHANGE existing data.** The 
-   simplest way to undedup existing data is to copy the deduped data to a new dataset 
-   or location that does not have dedup set. The data will be undeduped when written. 
-   The simplest way to undedup data and keep it in the same pool, zvol or dataset is 
-   as follows: disable dedup on the pool, zvol or dataset, then copy the data within 
-   the same pool, zvol or dataset. Once the copy is confirmed to be correct, the 
-   original data can optionally be deleted. If the pool does not have enough free space
-   for this procedure, then disable dedup, copy the data to another system or storage
-   device, check the copies are correct and safe, then delete the original data, and 
-   transfer back the copied data to the pool.
+**There is also no quick way to undedup data within a dataset once deduplication 
+is enabled, since disabling deduplication DOES NOT CHANGE existing data.** The procedure for removing deduplication varies depending upon the specific situation:
 
-   For these reasons, **think carefully before enabling dedup!** 
-   Deduplication is **usually not recommended unless the data contains a 
-   very substantial level of duplication, and the system (especially 
-   RAM) can cope acceptably with the additional burden**. In most cases, 
-   compression provides storage gains comparable to deduplication with 
-   less impact on performance.  This `article 
-   <http://constantin.glez.de/blog/2011/07/zfs-dedupe-or-not-dedupe>`_ 
-   provides a good description of the value versus cost considerations 
-   for deduplication. Unless a lot of RAM and a lot of highly duplicate 
-   data is available, **do not change the default deduplication setting 
-   of "Off"!**  
+* **To remove deduplication but keep the data:**  copy the deduped data to a new dataset 
+  or location that does not have dedup set. The data will be undeduped when written. 
 
-In %brand%, deduplication can be enabled during dataset creation, and 
-occurs at the time when files are read and written. The setting can 
-be changed at any time but will not affect data already written to 
-the pool.  If the deduplication method is set to *Verify*, ZFS will 
-do a byte-to-byte comparison when two blocks have the same hash signature 
-to make sure that the block contents are indeed identical. Since hash 
-collisions are extremely unblikely, *Verify* is usually not worth the 
-performance hit.
+* **To remove deduplication but keep the data in the same pool, zvol or dataset:**
+  disable deduplication on the pool, zvol or dataset, then copy the data within 
+  the same pool, zvol or dataset. The data will be copied without deduplication. 
+  Once the copy is confirmed to be correct, the original data can optionally be deleted.
+
+* **To remove deduplication but keep the data in the same pool, zvol or dataset,
+  if the pool does not have enough free capacity for the procedure above:**
+  disable deduplication, copy the data to a *different* system or storage device, 
+  check the copies are correct and safe, then delete the original data and finally, 
+  copy the copied data back to the original pool.
+
+In summary: **think carefully before enabling deduplication**. 
+Deduplication should only ever be considered if the data contains a 
+very substantial level of duplication, and the system (especially 
+RAM) can cope acceptably with the additional burden. In most cases, 
+compression provides storage gains comparable to deduplication with 
+less impact on performance.  This `article 
+<http://constantin.glez.de/blog/2011/07/zfs-dedupe-or-not-dedupe>`_ 
+provides a good description of the value versus cost considerations 
+for deduplication. 
 #endif freenas
 
 .. tip:: Deduplication is often considered when using a group of very
@@ -788,6 +789,16 @@ performance hit.
    overhead.
 
 #ifdef freenas
+.. _Deduplication method:
+
+Deduplication method
+--------------------
+If the deduplication method is set to *Verify*, ZFS will 
+do a byte-to-byte comparison when two blocks have the same hash signature 
+to make sure that the block contents are indeed identical. Since hash 
+collisions are extremely unblikely, *Verify* is usually not worth the 
+performance hit.
+
 .. _Deduplication RAM requirements:
 
 Deduplication RAM requirements
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -723,13 +723,14 @@ Deduplication
 Deduplication ("dedup") is a process where ZFS aims to store a single copy
 of any data that is repeated in different blocks or files, rather than storing 
 identical copies each time the data occurs. Deduplication 
-potentially allows a large number of very similar data to be stored very 
-efficiently in a smaller space, if there is a lot of identical data, but also places a 
-very heavy burden on RAM and on the CPU. In particular, dedup  
-significantly slows down data write performance and usually requires large (and at 
+potentially allows repeated and identical data to be stored very 
+efficiently in a smaller space, but also places a 
+very heavy burden on RAM and on the CPU. In particular, dedup 
+significantly reduces write performance and usually requires large (and at 
 times huge) amounts of RAM.  See :ref:`Deduplication RAM requirements` for details.
 Because of the demands it places on a system and the impact on performance, dedup
-is **not recommended** and should be left in the default setting ("off") in most situations.
+is **usually not recommended** and should be left in the default setting 
+("off") in most situations.
 
 .. warning::
 
@@ -769,9 +770,9 @@ In %brand%, deduplication can be enabled during dataset creation, and
 occurs at the time when files are read and written. The setting can 
 be changed at any time but will not affect data already written to 
 the pool.  If the deduplication method is set to *Verify*, ZFS will 
-do a byte-to-byte comparison when two blocks have the same signature 
-to make sure that the block contents are identical. Since hash 
-collisions are extremely rare, *Verify* is usually not worth the 
+do a byte-to-byte comparison when two blocks have the same hash signature 
+to make sure that the block contents are indeed identical. Since hash 
+collisions are extremely unblikely, *Verify* is usually not worth the 
 performance hit.
 #endif freenas
 
@@ -791,7 +792,7 @@ Deduplication RAM requirements
 
 It is often said that ZFS needs about 5 GB of RAM per terabyte of deduplicated 
 storage. However this may not be the case, depending on the data to be deduped. A 
-highly duplicated pool capabe of around 3x to 5x space saving may only need as little 
+highly duplicated pool capable of around 3x to 5x space saving may only need as little 
 as 1 or 2 GB of RAM per TB of data. The exact size required for deduplication data is 
 calculated from the output of the :command:`CLI` command :command:`zdb -U 
 /data/zfs/zpool.cache -D pool_name`, which states the total number of blocks in the 
@@ -806,18 +807,18 @@ ZFS uses RAM for many purposes. Therefore ZFS limits the amount of metadata
 main data cache ("ARC"). By default, DDTs and other file system metadata use up to 
 25% of ARC. Therefore if a system has 32 GB of RAM and 20 GB is available for ZFS, 
 ZFS would aim to apportion this 25% (5 GB) for metadata and 75% (15 GB) for file 
-data. The DDTs culd not rely on more than 5 GB of RAM (and perhaps less), which may 
+data. DDTs could not rely on more than 5 GB of RAM (and perhaps less), which may 
 only allow dedup of 1 TB in many cases.
 
-On systems with a large amount of RAM, metadata can use more RAM at the expense of 
-file cache data. For example: suppose that a 12 TB pool requires 30 GB for the DDTs 
+On systems with a large amount of RAM, where dedup is a possibility, more RAM can be
+reserved for metadata at the expense of file cache data. 
+For example: suppose that a 12 TB pool requires 30 GB for the DDTs 
 (2.5 GB per TB). On a system with 64 GB of RAM, the user could reserve 40 GB for 
 metadata and DDT, and leave the rest (24 GB) for the system, jails, and other file 
 cache. This would be done by creating a :ref:`Tunable` called :command:
 `vfs.zfs.arc_meta_limit` of type *loader*, and entering the amount of RAM to be used 
 for metadata, using G for GB or M for MB (in this case, 40G). The setting requires a 
 reboot to become activated. 
-
 #endif freenas
 
 .. index:: Compression
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -735,8 +735,8 @@ Deduplication can require **very large amounts of RAM** and substantial impacts 
 performance and CPU usage.  See :ref:`Deduplication RAM requirements` for details.
 
 Due to the demands that deduplication places on a system and the impact on performance,
-deduplication is **off by default**. Enabling deduplication without good reason is
-**not usually recommended** unless there are good reasons for doing so, the system has 
+deduplication is **off** by default and **should not be enabled** unless there
+are good reasons for doing so, the system has 
 sufficient RAM and CPU, and the performance impact has been tested on a sample of data and
 is acceptable.
 
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -814,9 +814,9 @@ data. DDTs could not rely on more than 5 GB of RAM (and perhaps less), which may
 only allow dedup of 1 TB in many cases.
 
 On systems with a large amount of RAM, where dedup is a possibility, more RAM can be
-reserved for metadata at the expense of file cache data. 
-For example: suppose that a 12 TB pool requires 30 GB for the DDTs 
-(2.5 GB per TB). On a system with 64 GB of RAM, the user could reserve 40 GB for 
+reserved for metadata at the expense of file cache data. For example: suppose that 
+deduping 12 TB of undeduped data requires 30 GB of RAM for the DDTs (2.5 GB per TB). 
+On a system with 64 GB of RAM, the user could reserve 40 GB for 
 metadata and DDT, and leave the rest (24 GB) for the system, jails, and other file 
 cache. This would be done by creating a :ref:`Tunable` called :command:
 `vfs.zfs.arc_meta_limit` of type *loader*, and entering the amount of RAM to be used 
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -713,53 +713,64 @@ existing datasets to close them--the remaining
...
-   from :ref:`Shell`. However, any data that has already been
-   deduplicated will not be un-deduplicated. Only newly stored data
-   after the property change will not be deduplicated. The only way to
-   remove existing deduplicated data is to copy all of the data off of
-   the dataset, set the property to off, then copy the data back in
-   again. Alternately, create a new dataset with
-   :guilabel:`ZFS Deduplication` left disabled, copy the data to the
-   new dataset, and destroy the original dataset.
+Deduplication (often just called "dedup") is the process of storing 
+just one copy of any data that is found to be the same in different
```
wblock: This sentence starts with a halting aside and then has a really confusing construct.  This whole paragraph repeatedly talks about files, implying that dedup works on files rather than blocks, which is misleading. Dedup could also work on binary data in zvols, for example.  And we often have people mistakenly thinking that dedup works at the file level.

userguide/storage.rst
```diff
@@ -713,53 +713,64 @@ existing datasets to close them--the remaining
...
+   so that the deduplicated data in the pool can be checked during import. 
+   If the system does not have the needed RAM, it will panic. The only 
+   solution then is to add more RAM or recreate the pool.
+
+   **There is also no quick way to undedup data 
+   within a dataset once deduplication is enabled**, as disabling 
+   deduplication has **NO EFFECT** on existing data. The simplest way to 
+   undedup existing data is to copy the deduped files to a new undeduped 
+   dataset or location in the pool or elsewhere, delete the original 
+   data, and (if necessary) copy back the undeduped copies of the data.
```
wblock: Lots of asides in this paragraph, which interrupt the flow of sentences and distract from the primary point.  Move asides to separate sentences.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
 
+#ifdef freenas
+.. _Deduplication RAM requirements:
+
+Deduplication RAM requirements
+------------------------------
+
+A generally quoted rule of thumb is 5 GB of RAM per terabyte of 
+deduplicated storage. However, if the pool is highly duplicated 
+(around 3x to 5x space saving) this can drop to as low as 1 GB to 2
```
wblock: The rest of these changes also have lots of asides.  Please avoid that.  Even a comma-based aside is better than parentheses, like this one, but having short, declarative sentences without pauses is best.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+deduplicated storage. However, if the pool is highly duplicated 
+(around 3x to 5x space saving) this can drop to as low as 1 GB to 2 
+GB of RAM per TB of deduplicated storage. The exact size required for 
+deduplication data is found by using the :command:`CLI` command 
+:command:`zdb -U /data/zfs/zpool.cache -D pool_name` to identify the 
+total number of blocks in the pool, and the number of bytes of RAM 
+required per block (shown as *"{number} in core"* in the output). The 
+command  :command:`zdb -U /data/zfs/zpool.cache -S pool_name` 
+displays an estimate of the storage saving if deduplication is 
+applied to a dataset. Note that these commands may take a long time
```
wblock: Try to avoid using the word "note" to introduce a sentence, as the existence of the sentence means the reader is supposed to note it anyway.

stilez:
@wblock  @miker - edits done to address comments by wblock. Hopefully somewhat happier now :)

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
-   :guilabel:`ZFS Deduplication` left disabled, copy the data to the
-   new dataset, and destroy the original dataset.
+#ifdef freenas
+.. index:: Deduplication
+.. _Deduplication:
+
+Deduplication
+^^^^^^^^^^^^^
+
+Deduplication ("dedup") is a process where ZFS aims to store a single copy
```
miker: We should get rid of ("dedup") especially in this technical doc, we should use the full word deduplication.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+of any data that is repeated in different blocks or files, rather than storing
```
miker: Condense this down to "a single copy of block data" From what I know, ZFS only does deduplication of block data, not file deduplication at all.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+#ifdef freenas
+.. index:: Deduplication
+.. _Deduplication:
+
+Deduplication
+^^^^^^^^^^^^^
+
+Deduplication ("dedup") is a process where ZFS aims to store a single copy
+of any data that is repeated in different blocks or files, rather than storing 
+identical copies each time the data occurs. Deduplication
```
miker: Need to rework this to describe how ZFS computes a cryptographic hash of block data.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+
+Deduplication
+^^^^^^^^^^^^^
+
+Deduplication ("dedup") is a process where ZFS aims to store a single copy
+of any data that is repeated in different blocks or files, rather than storing 
+identical copies each time the data occurs. Deduplication 
+potentially allows repeated and identical data to be stored very 
+efficiently in a smaller space, but also places a 
+very heavy burden on RAM and on the CPU. In particular, dedup
```
miker: Rework the sentence to get rid of 'potentially'. That makes it sound like we aren't sure it works. Should stick to confirmed information. Get rid of all ambiguous "data" terms. Refer to block data specifically. 
Remove "in particular" and "dedup" start with the full word Deduplication. We aren't paying by the character ;) Rework the sentence to describe how the data is stored. This should hopefully allow the removal of ", but" split into a separate sentence if need be.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+^^^^^^^^^^^^^
+
+Deduplication ("dedup") is a process where ZFS aims to store a single copy
+of any data that is repeated in different blocks or files, rather than storing 
+identical copies each time the data occurs. Deduplication 
+potentially allows repeated and identical data to be stored very 
+efficiently in a smaller space, but also places a 
+very heavy burden on RAM and on the CPU. In particular, dedup 
+significantly reduces write performance and usually requires large (and at 
+times huge) amounts of RAM.  See :ref:`Deduplication RAM requirements` for details.
```
miker: Get rid of usually and the aside in the parens. State the amount of RAM required, which is then followed up by the link to the RAM requirements.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+Because of the demands that dedup places on a system and the impact on performance, 
+enabling dedup is **usually not recommended** and the setting should be left in
+the default setting ("off") in most situations.
+
```
miker: remove "Because of the" Something like:
Due to the demands that deduplication places on a system and the impacts on performance, enabling deduplication is **not recommended**. Deduplication is off by default and should not be enabled without good reason.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+   requires. The DDTs (dedup tables) also compete with other data for RAM,
+   and can be pushed out of RAM in some cases. When the system starts 
+   storing DDTs on disk because they no longer fit into RAM or have been
+   pushed out of RAM, performance craters.
+   
+   Importing an unclean deduplicated pool can also require large amounts of RAM, 
+   so that the deduplicated data in the pool can be checked during import. 
+   If the system does not have the needed RAM, it will panic. The only 
+   solution then is to add more RAM or recreate the pool.
+
```
miker: Turn this into its own paragraph. Reword this to state the what and why. Remove "it" in the rewrite.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+   **There is also no quick way to undedup data within a dataset once deduplication 
+   is enabled, since disabling deduplication DOES NOT CHANGE existing data.** The 
+   simplest way to undedup existing data is to copy the deduped data to a new dataset 
+   or location that does not have dedup set. The data will be undeduped when written.
```
miker: This needs to be rewritten. Remove things like "There is also no" State the facts.
Get rid of words like "undedup" For the "There is" sentence, something like:
Enabling deduplication on a dataset cannot easily be undone. To remove deduplication the data needs to be written to a new dataset or location that does not have deduplication enabled.
Possibly move "Disabling deduplication will not remove the deduplication of block data that has been deduplicated." to a warning. That sentence my need to be reworded.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+   simplest way to undedup existing data is to copy the deduped data to a new dataset 
+   or location that does not have dedup set. The data will be undeduped when written. 
+   The simplest way to undedup data and keep it in the same pool, zvol or dataset is 
+   as follows: disable dedup on the pool, zvol or dataset, then copy the data within 
+   the same pool, zvol or dataset. Once the copy is confirmed to be correct, the 
+   original data can optionally be deleted. If the pool does not have enough free space
+   for this procedure, then disable dedup, copy the data to another system or storage
+   device, check the copies are correct and safe, then delete the original data, and 
+   transfer back the copied data to the pool.
+
```
miker: Remove "The simplest way to undedup" Start as a statement. "To remove deduplication but keep the same"
remove "as follows" and just give the procedure.
Rewrite this section with the above in mind.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+   very substantial level of duplication, and the system (especially 
+   RAM) can cope acceptably with the additional burden**. In most cases, 
+   compression provides storage gains comparable to deduplication with 
+   less impact on performance.  This `article 
+   <http://constantin.glez.de/blog/2011/07/zfs-dedupe-or-not-dedupe>`_ 
+   provides a good description of the value versus cost considerations 
+   for deduplication. Unless a lot of RAM and a lot of highly duplicate 
+   data is available, **do not change the default deduplication setting 
+   of "Off"!**  
+
```
miker: Get rid of "For these reasons" Rewrite to get rid of the aside. 
Rewrite to use directions/statements of fact.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+In %brand%, deduplication can be enabled during dataset creation, and 
+occurs at the time when files are read and written. The setting can 
+be changed at any time but will not affect data already written to 
+the pool.  If the deduplication method is set to *Verify*, ZFS will 
+do a byte-to-byte comparison when two blocks have the same hash signature 
+to make sure that the block contents are indeed identical. Since hash 
+collisions are extremely unblikely, *Verify* is usually not worth the 
+performance hit.
```
miker: Is this repeated information? I'd say combine this information into the sections above where applicable. After that turn only what's different should be in this final paragraph.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+deduplication data is found by using the :command:`CLI` command 
+:command:`zdb -U /data/zfs/zpool.cache -D pool_name` to identify the 
+total number of blocks in the pool, and the number of bytes of RAM 
+required per block (shown as *"{number} in core"* in the output). The 
+command  :command:`zdb -U /data/zfs/zpool.cache -S pool_name` 
+displays an estimate of the storage saving if deduplication is 
+applied to a dataset. Note that these commands may take a long time 
+to run because they scan the entire pool to produce the results. 
+
+There is also another RAM limitation affecting dedup.  Because ZFS
```
miker: Is the "There is also" sentence needed? The following sentence should explain the fact there is another limitation.
Never start sentences with "because"

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+total number of blocks in the pool, and the number of bytes of RAM 
+required per block (shown as *"{number} in core"* in the output). The 
+command  :command:`zdb -U /data/zfs/zpool.cache -S pool_name` 
+displays an estimate of the storage saving if deduplication is 
+applied to a dataset. Note that these commands may take a long time 
+to run because they scan the entire pool to produce the results. 
+
+There is also another RAM limitation affecting dedup.  Because ZFS 
+uses RAM for many caching activities, it limits the amount of metadata
+(including DDT data) for which it attempts to reserves space in RAM. By
```
miker: Get rid of "it" and the aside. Rework the sentence into 2 sentences if need be.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+default, DDTs and other file system metadata can rely on up to 25% of
+the total RAM available to ZFS (known as ARC), and the balance is used 
+for  other caching. Therefore if the system has 32 GB of RAM of which 
+20 GB is available for ZFS, ZFS will aim to apportion this 25% (5 GB) 
+for metadata and 75% (15 GB) for file data. The DDTs can rely on not more 
+than 5 GB of RAM (and perhaps less), which may only allow dedup of 
+1 TB in many cases.
+
```
miker: Lots of asides here. Rework the sentences to get rid of those.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+If the system has a large amount of RAM, metadata can use more RAM, at 
+the expense of file cache data. For example: suppose that a 12 TB pool with
+a dedup ratio of 3x space saving is found to require 30 GB total for the DDT 
+(2.5 GB per TB). If the system has 64 GB of RAM then the user could reserve 
+40 GB for metadata and DDT, and leave the rest (24 GB) for the system, jails, and
+other file cache.  This would  be done by creating a :ref:`Tunable` called 
+:command:`vfs.zfs.arc_meta_limit` of type *loader*, and entering the amount 
+of RAM to be used for metadata. The value can be abbreviated using G for GB
+or M for MB (in this case, 40G). The setting requires a reboot to become activated.
+
```
miker: Rework the sentence starting with "If" to remove the 2 commas. Turning the sentence into statements instead of possibilities should help. "On systems with large amounts of RAM, metadata will use RAM at the expense of file cache data." something like that.
The rest of the paragraph needs to be rewritten as factual instruction instead of a story.

userguide/storage.rst
```diff
@@ -770,6 +781,47 @@ usually not worth the performance hit.
...
+#endif freenas
```
miker: Thanks for the contribution! Apologies for the delay in the review.

userguide/storage.rst
stilez: Not happy with this:  
This isn't just a "technical document". It's a user guide. People who seek help or want to read more will inevitably see the term "dedup" used, as well as/in preference to "deduplication". So we should at least *mention* the abbreviation in passing, since it's used **everywhere** in the real world when discussing the feature.

userguide/storage.rst
stilez: For similar reason I'm not happy about the requested change to refer only to "block data". Yes technically ZFS dedups on blocks. No, end-users will not necessarily know what that means, nor should they be expected to if not using a zvol, why confuse them? If the user is storing files, then it's deduping files, albeit by means of deduping the blocks that constitute the files. *(A bit like saying the OS doesn't read a file, it's reading sectors that make up a file. Technically true, in practice pointless for "guide" purposes)*

userguide/storage.rst
stilez: For similar reasons I'm not happy again: this is a guide that has to cater to all users. Why exactly does an ordinary user need a tutorial about the technical mechanics of how dedup is done "under the radar", or a tutorial on cryptographic hashing, beyond being told that ZFS identifies repeated data and dedup's it? They don't, for this kind of guide. The only place this is relevant is when choosing dedup algorithm where the guide already explains all that a user will need to know, to make a choice. (If not, fix it there not here).

userguide/storage.rst
stilez: Again, for a user storing files, what they see is the file being deduped, they **never** see "blocks" being deduped, nor do we need to dive into files being stored in blocks and blocks being the actual basis of deduplication. It just doesn't add anything, pragmatically, if the user knows this subtlety having not known it before.  (See previous comment). It's like insisting a user knows about unix pipes or 512B/4K disk sectors when discussing file backup, No actual user benefit other than engineering satisfaction :)

Did the rest.

userguide/storage.rst
stilez: 1) done

2) You can't "state the amount of RAM required - it depends totally on the extent of duplicate data and the structure/size of the pool. Some systems will need 5GB per TB, some will need 1GB per TB. Mine currently has a "dedup" and "nondedup" dataset in the pool, to bypass dedup processing for data I know won't be likely to deduplicate - the overall pool requirement is about 1.6GB per TB. You also need to consider how much RAM will be reserved for ARC and within that how much ARC is reserved for metadata - a user with 32GB RAM considering dedup might quite reasonably buy another 32GB and dedicate it 100% to metadata, while a different user might keep the 25% default ratio. 

**TL;DR you can't do what you're asking in the 2nd sentence.** All you can say here, is that dedup needs a lot (and in some cases a huge amount) of RAM and see below for more exact details how to work out the requirement for *your* data, and what you can do to trim it.

userguide/storage.rst
stilez: Done. Reworded slightly, might need review.

userguide/storage.rst
```diff
@@ -713,53 +713,70 @@ existing datasets to close them--the remaining
...
+the default setting ("off") in most situations.
+
+.. warning::
+
+   The more data written to a deduplicated dataset, the more RAM it 
+   requires. The DDTs (dedup tables) also compete with other data for RAM,
+   and can be pushed out of RAM in some cases. When the system starts 
+   storing DDTs on disk because they no longer fit into RAM or have been
+   pushed out of RAM, performance craters.
+
```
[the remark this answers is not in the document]
stilez: Done - is rewrite ok?

userguide/storage.rst
stilez: Done reword. 
Does it really need a paragraph split, as it follows directly from the previous sentence?

userguide/storage.rst
stilez: Rewritten as bullet list, to cover the different scenarios

userguide/storage.rst
stilez: Tried rewording it in line with comments. Please review.

userguide/storage.rst
stilez: This para covers 2 distinct points: 

1)  dedup is set at creation and can be modified afterwards. (a general point about how it works on *NAS)
2) the dedup algorithm can be changed and verfiy used, but isn't that helpful normally. (a specific choice of parameter)

I've split these and moved to better places, merging in as possible. The first is moved right near the start of the section, the second is left further down.

userguide/storage.rst
wblock: I think it is okay to mention it, then use the short term after that: "Deduplication, or "dedup", is where..."

userguide/storage.rst
wblock: The problem is that file-level dedup is almost always what leads users to think they need it.  And then they regret it.  I'm in favor of simplifying things, but not to the point that it hides a crucial fact.  And the blocks used by two files are not guaranteed to be identical.  One could be offset due to block sub-allocation or something.

userguide/storage.rst
wblock: I think this can be simplified to explain that the checksum of a block is calculated when the block is saved, then dedup just compares those checksums.

stilez: @wblock - if it helped inform a user's choices I'd be all for it, but I just don't see any relevance yet. Maybe I'm missing something? Meaning - what impact does this knowledge (or its lack), concerning whether dedup is done via hashes, byte compare, or magic pixie dust, have on a user or their NAS decisions **here**? *It just doesn't impact users*. If the speed, or resource use, impacts users then we can say that, but even then, a knowledge of the technique ZFS uses "under the hood" has no impact, so it seems irrelevant. Are we sure there's a tangible need for this, that it's not just something like "a user just ought to know this detail otherwise how will they appreciate it" or similar?

userguide/storage.rst
stilez: In that case the issue is the warnings, not the belief in file-level dedup vs block level dedup. We aren't warning of the right things - which is that file level dedup may not realise expected benefits since it doesan't work that way. We don't need to go into how it actually does work, to make that point with more force.

We might instead warn that *dedup does not always deduplicate even apparently-identical files, as it works at the level of blocks rather than files.  **It is imperative to actually check the benefits of deduplication and its impact on performance, even with highly duplicated data. Do not assume any benefits might be realised from enabling this setting, unless tests have actually confirmed this to be the case**.*

Would something like that work? The whole issue of block vs file is a red herring it seems - the user doesn't need to know exactly how it works. They need to know that what they might assume, isn't so, and not to assume it. A brief mention of blocks (as above) is sufficient to underpin that point.

userguide/storage.rst
stilez: Nice phrasing!

userguide/storage.rst
stilez: In that case the issue is the warnings. It's not the lack of education about file-level dedup vs block level dedup. The problem is, we're not warning of the right things, or not warning of them adequately - namely that file level dedup may not realise the naively expected benefits since ZFS doesn't actually work that way. But we don't need to go in depth about how ZFS actually does work, to make that point with necessary force.

We might instead warn that: *The deduplication function does not always deduplicate even apparently-identical files, because it works at the level of data blocks on disk, and not on individual files.  **Therefore it is imperative to actually check the benefits of deduplication and its impact on performance, even with highly duplicated data, before assuming that deduplication will provide any benefits whatsoever. The number of cases where deduplication is actually useful, is small. Do not assume any benefits will be realised from enabling this setting, in any situation, unless your own tests have actually confirmed this to be the case**.*

Would something like that work? The whole issue of block vs file is a red herring it seems - the user doesn't need to know exactly how it works. They need to know that what they might assume, isn't so, and not to assume it. A brief mention of blocks (as above) is sufficient to underpin that point.

userguide/storage.rst
stilez: This para covers 2 distinct points: 

1)  dedup is set at creation and can be modified afterwards. (a general point about how it works on *NAS)
2) the dedup algorithm can be changed and verfiy used, but isn't that helpful normally. (a specific choice of parameter)

I've split these and moved to better places, merging in as possible. The first is moved right near the start of the section, the second is left further down.

I notice that some of the info is omitted for TrueNAS but other info on dedup is left in. Can you check if the info left in/out is correct?

userguide/storage.rst
stilez: updated, please re-review

userguide/storage.rst
stilez: updated, please re-review

userguide/storage.rst
stilez: updated, please re-review

userguide/storage.rst
stilez: updated, please re-review

userguide/storage.rst
stilez: Nice phrasing! Updated.

userguide/storage.rst
stilez: @wblock - if it helped inform a user's choices I'd be all for it, but I just don't see any relevance yet. Maybe I'm missing something? Meaning - what impact does this knowledge (or its lack), concerning whether dedup is done via hashes, byte compare, or magic pixie dust, have on a user or their NAS decisions **here**? *It just doesn't impact users*. If the speed, or resource use, impacts users then we can say that, but even then, a knowledge of the technique ZFS uses "under the hood" has no impact, so it seems irrelevant. Are we sure there's a tangible need for this, that it's not just something like "a user just ought to know this detail otherwise how will they appreciate it" or similar?

userguide/storage.rst
stilez: Again, for a user storing files, what they see is the file being deduped, they **never** see "blocks" being deduped, nor do we need to dive into files being stored in blocks and blocks being the actual basis of deduplication. It just doesn't add anything, pragmatically, if the user knows this subtlety having not known it before.  (See previous comment). It's like insisting a user knows about unix pipes or 512B/4K disk sectors when discussing file backup, No actual user benefit other than engineering satisfaction :)

Did the rest.

userguide/storage.rst
stilez: 1) done

2) You can't "state the amount of RAM required" - it totally depends on the extent of duplicate data and the structure/size of the pool. Some systems will need 5GB or more per TB, some will need <1GB per TB. Mine currently has a "dedup" and "nondedup" dataset in the pool, to bypass dedup processing for data I know won't be likely to deduplicate - the overall pool requirement is about 1.6GB per TB. You also need to consider how much RAM will be reserved for ARC and within that how much ARC is reserved for metadata - a user with 32GB RAM considering dedup might quite reasonably buy another 32GB and dedicate it 100% to metadata, while a different user might keep the 25% default ratio. 

**TL;DR you can't do what you're asking in the 2nd sentence.** All you can say here, is that dedup needs a lot (and in some cases a huge amount) of RAM and see below for more exact details how to work out the requirement for *your* data, and what you can do to trim it.

userguide/storage.rst
stilez: Done. Reworded slightly, might need review.

userguide/storage.rst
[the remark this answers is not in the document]
stilez: Done - is rewrite ok?

userguide/storage.rst
stilez: Done reword.
Does it really need a paragraph split, as it follows directly from the previous sentence?

userguide/storage.rst
stilez: Done.
Also reformatted as a bullet list to cover the different scenarios

userguide/storage.rst
stilez: Tried rewording it in line with comments. Please review.

userguide/storage.rst
stilez: This para covers 2 distinct points: 

1)  dedup is set at creation and can be modified afterwards. (a general point about how it works on *NAS)
2) the dedup algorithm can be changed and verfiy used, but isn't that helpful normally. (a specific choice of parameter)

I've split these and moved to better places, merging in as possible. The first is moved right near the start of the section, the second is left further down.

I notice that some of the info is omitted for TrueNAS but other info on dedup is left in. Can you check if the info left in/out is correct?

userguide/storage.rst
stilez: Text has been updated - please re-review current version

userguide/storage.rst
stilez: Text has been updated - please re-review current version

userguide/storage.rst
stilez: Text has been updated - please re-review current version

userguide/storage.rst
stilez: Welcome. Hope the response comments are helpful. Please re-review

miker:
After discussing with Dru, it was decided that this change will not be made as we do not recommend dedup at all. Compression is currently the preferred way to save space if desired.

PR #68, opened 2018-04-09
Update logout.rst:

Mrt134:
Update text to reflect the logout process in the new UI.
HTML build test: no issues.

### DIFF
--- a/userguide/logout.rst
+++ b/userguide/logout.rst
@@ -4,7 +4,11 @@
 Log Out
 =======
 
-Click the :guilabel:`Log Out` entry in the %brand% GUI to log out.
+Click :guilabel:`Settings` in the upper right corner then
+:guilabel:`Log Out` in the %brand% GUI to log out.
 
-After logging out, a message appears with a link to log back in. When
-logging back in, the *root* password is required.
+A popup window confirms the choice. Click :guilabel:`Ok` and the
+system continues the log out process. The window returns to the %brand%
+log in screen when finished.
+
+When logging back in, the *root* credentials are required.


userguide/logout.rst
```diff
@@ -4,7 +4,11 @@
 Log Out
 =======
 
-Click the :guilabel:`Log Out` entry in the %brand% GUI to log out.
+Click :guilabel:`Settings` in the upper right corner then
```
wblock: "Upper right" is enough.

userguide/logout.rst
```diff
@@ -4,7 +4,11 @@
...
+:guilabel:`Log Out` in the %brand% GUI to log out.
 
-After logging out, a message appears with a link to log back in. When
-logging back in, the *root* password is required.
+A popup window confirms the choice. Click :guilabel:`Ok` and the
```
wblock: This all seems very wordy and roundabout.  "The system continues the log out process." is not very clear.
How about "Confirm the choice to log out."?

userguide/logout.rst
```diff
@@ -4,7 +4,11 @@
...
+system continues the log out process. The window returns to the %brand%
+log in screen when finished.
+
+When logging back in, the *root* credentials are required.
```
wblock: What are the "root credentials"?  We don't use that wording elsewhere, do we?  Do you mean a password, SSH key, biometric information?

We can/should probably not mention that at all.

userguide/logout.rst
```diff
@@ -4,7 +4,11 @@
 Log Out
 =======
 
-Click the :guilabel:`Log Out` entry in the %brand% GUI to log out.
+Click :guilabel:`Settings` in the upper right corner then
+:guilabel:`Log Out` in the %brand% GUI to log out.
```
wblock: "in the %brand% GUI" ...as opposed to what?

PR #69, opened 2018-04-10
Initial draft of reporting.rst against the new GUI.

a-stjohn:


dlavigne:
Aaron: I'm merging in this request, but it is a mix of text improvements (which also apply to the legacy Guide) and text edits specific to the new UI. Please also create a PR against master that only contains the text improvements.

PR #126, opened 2018-05-31
Update tabularcolumns entries to fix PDF builds:

Mrt134:
- errors with these .. tabularcolumns entries was erroring out pdf builds.
- Search through angulargui .rst files and fix tabularcolumn entries until the PDF builds properly again.
- Note: there are still lots of warnings, but the builder will finish now.


dlavigne:
Does this also affect master?

PR #132, opened 2018-06-07
Update services.rst in angulargui.

sambonham:


dlavigne:
Don't forget to port doc improvments to master.

PR #140, opened 2018-06-12
Misc updates to AFP and SMB sharing docs

seanm:
Notably:
- note that Time Machine does not yet work on SMB
- use current name of "macOS" where appropriate
- other clarifications after reading this for the first time and using it to try FreeNAS for the first time

### DIFF
--- a/userguide/sharing.rst
+++ b/userguide/sharing.rst
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
 
 These types of shares and services are available:
 
-* :ref:`AFP <Apple (AFP) Shares>`: Apple File Protocol shares are
-  often used when the client computers all run macOS. Apple has
-  slowly shifted to preferring :ref:`SMB <Windows (SMB) Shares>` for
-  modern networks, although Time Machine still requires AFP.
+* :ref:`AFP <Apple (AFP) Shares>`: Apple Filing Protocol shares are
+  used when the client computers all run macOS. Apple has deprecated
+  AFP in favour of :ref:`SMB <Windows (SMB) Shares>` and using AFP
+  in modern networks is no longer recommanded (except for providing a
+  Time Machine, for which %brand% currently only supports AFP, not
+  yet SMB).
 
 * :ref:`Unix (NFS) <Unix (NFS) Shares>`: Network File System shares
-  are accessible from Mac OS X, Linux, BSD, and the professional and
+  are accessible from macOS, Linux, BSD, and the professional and
   enterprise versions (but not the home editions) of Windows. This can
   be are a good choice when the client computers do not all run the
   same operating system but NFS client software is available for all
@@ -53,10 +55,10 @@ These types of shares and services are available:
 
 * :ref:`SMB <Windows (SMB) Shares>`: Server Message Block shares, also
   known as Common Internet File System (CIFS) shares, are accessible
-  by Windows, Mac OS X, Linux, and BSD computers. Access is slower
+  by Windows, macOS, Linux, and BSD computers. Access is slower
   than an NFS share due to the single-threaded design of Samba. SMB
-  provides more configuration options than NFS and is a good choice
-  on a network for Windows systems. However, it is a poor choice if
+  provides more configuration options than NFS and is the best choice
+  on a network for Windows or Mac systems. However, it is a poor choice if
   the CPU on the %brand% system is limited; if the CPU is maxed out,
   upgrade the CPU or consider another type of share.
 
@@ -197,8 +199,8 @@ information given when the share was created.
    |                              |               |          | automounting or volumes created by a preexec script.                                                          |
    |                              |               |          |                                                                                                               |
    +------------------------------+---------------+----------+---------------------------------------------------------------------------------------------------------------+
-   | AFP3 UNIX Privs              | checkbox      | ✓        | Enable Unix privileges supported by OSX 10.5 and higher. Do not enable this if the network contains           |
-   |                              |               |          | macOS 10.4 clients or lower as they do not support this feature.                                              |
+   | AFP3 UNIX Privs              | checkbox      | ✓        | Enable Unix privileges supported by Mac OS X 10.5 and higher. Do not enable this if the network contains           |
+   |                              |               |          | Mac OS X 10.4 clients or lower as they do not support this feature.                                              |
    |                              |               |          |                                                                                                               |
    +------------------------------+---------------+----------+---------------------------------------------------------------------------------------------------------------+
    | Default file permission      | checkboxes    | ✓        | Only works with Unix ACLs. New files created on the share are set with the selected permissions.              |
@@ -285,7 +287,7 @@ the AFP service so the share is immediately available. The new share
 is also added as an entry to
 :menuselection:`Sharing --> Apple (AFP)`.
 
-macOS users can connect to the guest AFP share by clicking
+macOS users can connect to the guest AFP share from the Finder by clicking
 :menuselection:`Go --> Connect to Server`. In the example shown in
 :numref:`Figure %s <afp_connect_server_fig>`,
 the user has entered *afp://* followed by the IP address of the
@@ -314,10 +316,11 @@ the :guilabel:`Shared` sidebar.
 Creating Authenticated and Time Machine Shares
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-macOS includes the Time Machine application which can be used to
-schedule automatic backups.  In this configuration example, a Time
+macOS includes the Time Machine feature which can be used to
+perform automatic backups.  In this configuration example, a Time
 Machine user will be configured to backup to an AFP share on a
-%brand% system. Creating a separate Time Machine
+%brand% system. (%brand% does not yet support Time Machine over SMB.)
+Creating a separate Time Machine
 share for each user that will be using Time Machine to backup their
 macOS system to %brand% is recommended. The process for creating
 an authenticated share for a user is the same as creating a
@@ -755,7 +758,7 @@ best results, use :ref:`Windows (SMB) Shares`.
 From macOS
 ^^^^^^^^^^^^^
 
-To mount the NFS volume from a macOS client, click on
+To mount the NFS volume from a macOS client, from the Finder click on
 :menuselection:`Go --> Connect to Server`.
 In the :guilabel:`Server Address` field, enter *nfs://* followed by
 the IP address of the %brand% system and the name of the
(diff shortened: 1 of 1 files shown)


userguide/sharing.rst
```diff
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
 
 These types of shares and services are available:
 
-* :ref:`AFP <Apple (AFP) Shares>`: Apple File Protocol shares are
-  often used when the client computers all run macOS. Apple has
-  slowly shifted to preferring :ref:`SMB <Windows (SMB) Shares>` for
-  modern networks, although Time Machine still requires AFP.
+* :ref:`AFP <Apple (AFP) Shares>`: Apple Filing Protocol shares are
+  used when the client computers all run macOS. Apple has deprecated
+  AFP in favour of :ref:`SMB <Windows (SMB) Shares>` and using AFP
```
wblock: Please use American English spelling.

userguide/sharing.rst
```diff
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
...
+  in modern networks is no longer recommanded (except for providing a
```
wblock: Typo: recommended.

userguide/sharing.rst
```diff
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
...
+  Time Machine, for which %brand% currently only supports AFP, not
+  yet SMB).
```
wblock: This sentence is very confusing and probably needs to be split into at least two sentences.

userguide/sharing.rst
```diff
@@ -755,7 +758,7 @@ best results, use :ref:`Windows (SMB) Shares`.
 From macOS
 ^^^^^^^^^^^^^
 
-To mount the NFS volume from a macOS client, click on
+To mount the NFS volume from a macOS client, from the Finder click on
```
wblock: "from a macOS client, from the Finder" sounds redundant, even if it isn't.  Sentence probably could be rearranged to avoid that.

userguide/sharing.rst
```diff
@@ -1379,7 +1382,7 @@ the administrator to control access to data, provide appropriate
 permissions to that data, and to determine who accesses and modifies
 stored data. A Windows domain controller is not needed for
 authenticated SMB shares, which means that additional licensing costs
-are not required. However, since there is no domain controller to
+are not required. However, when there is no domain controller to
```
wblock: Maybe "because" would capture the meaning better.

userguide/sharing.rst
```diff
@@ -1379,7 +1382,7 @@ the administrator to control access to data, provide appropriate
...
 provide authentication for the network, each user account needs to be
```
wblock: s/needs to/must/

userguide/sharing.rst
```diff
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
 
 These types of shares and services are available:
 
-* :ref:`AFP <Apple (AFP) Shares>`: Apple File Protocol shares are
-  often used when the client computers all run macOS. Apple has
-  slowly shifted to preferring :ref:`SMB <Windows (SMB) Shares>` for
-  modern networks, although Time Machine still requires AFP.
+* :ref:`AFP <Apple (AFP) Shares>`: Apple Filing Protocol shares are
+  used when the client computers all run macOS. Apple has deprecated
```
wblock: This needs a qualifier, like "often", or maybe "only".  Even that is too absolute.  Something like "...are generally used only when legacy client computers all run macOS."

userguide/sharing.rst
```diff
@@ -34,13 +34,15 @@ system to meet the requirements of the network.
...
+  AFP in favour of :ref:`SMB <Windows (SMB) Shares>` and using AFP
+  in modern networks is no longer recommanded (except for providing a
+  Time Machine, for which %brand% currently only supports AFP, not
+  yet SMB).
 
 * :ref:`Unix (NFS) <Unix (NFS) Shares>`: Network File System shares
-  are accessible from Mac OS X, Linux, BSD, and the professional and
+  are accessible from macOS, Linux, BSD, and the professional and
   enterprise versions (but not the home editions) of Windows. This can
   be are a good choice when the client computers do not all run the
```
wblock: s/be are/be/

userguide/sharing.rst
```diff
@@ -53,10 +55,10 @@ These types of shares and services are available:
 
 * :ref:`SMB <Windows (SMB) Shares>`: Server Message Block shares, also
   known as Common Internet File System (CIFS) shares, are accessible
-  by Windows, Mac OS X, Linux, and BSD computers. Access is slower
+  by Windows, macOS, Linux, and BSD computers. Access is slower
   than an NFS share due to the single-threaded design of Samba. SMB
-  provides more configuration options than NFS and is a good choice
-  on a network for Windows systems. However, it is a poor choice if
+  provides more configuration options than NFS and is the best choice
```
wblock: "best" is subjective.  I think the original "a good choice" is correct.  NFS is much more performant if the client software is available.

userguide/sharing.rst
```diff
@@ -53,10 +55,10 @@ These types of shares and services are available:
...
+  on a network for Windows or Mac systems. However, it is a poor choice if
   the CPU on the %brand% system is limited; if the CPU is maxed out,
```
wblock: Split the sentence at the semicolon, replacing it with a period.

userguide/sharing.rst
```diff
@@ -53,10 +55,10 @@ These types of shares and services are available:
...
   upgrade the CPU or consider another type of share.
```
wblock: s/the CPU/it/
s/another/a different/

userguide/sharing.rst
```diff
@@ -197,8 +199,8 @@ information given when the share was created.
    |                              |               |          | automounting or volumes created by a preexec script.                                                          |
    |                              |               |          |                                                                                                               |
    +------------------------------+---------------+----------+---------------------------------------------------------------------------------------------------------------+
-   | AFP3 UNIX Privs              | checkbox      | ✓        | Enable Unix privileges supported by OSX 10.5 and higher. Do not enable this if the network contains           |
-   |                              |               |          | macOS 10.4 clients or lower as they do not support this feature.                                              |
+   | AFP3 UNIX Privs              | checkbox      | ✓        | Enable Unix privileges supported by Mac OS X 10.5 and higher. Do not enable this if the network contains           |
```
wblock: s/Do not enable this if/Do not enable if/
s/contains/has/

userguide/sharing.rst
```diff
@@ -197,8 +199,8 @@ information given when the share was created.
...
+   |                              |               |          | Mac OS X 10.4 clients or lower as they do not support this feature.                                              |
```
wblock: s/10.4 clients or lower/10.4 or lower clients/

userguide/sharing.rst
```diff
@@ -314,10 +316,11 @@ the :guilabel:`Shared` sidebar.
 Creating Authenticated and Time Machine Shares
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-macOS includes the Time Machine application which can be used to
-schedule automatic backups.  In this configuration example, a Time
+macOS includes the Time Machine feature which can be used to
+perform automatic backups.  In this configuration example, a Time
 Machine user will be configured to backup to an AFP share on a
```
wblock: s/will be/is/
s/backup/back up/

userguide/sharing.rst
```diff
@@ -314,10 +316,11 @@ the :guilabel:`Shared` sidebar.
...
-%brand% system. Creating a separate Time Machine
+%brand% system. (%brand% does not yet support Time Machine over SMB.)
+Creating a separate Time Machine
```
wblock: "A separate Time Machine share is recommended for each user backing up their system."

userguide/sharing.rst
seanm: OK.

userguide/sharing.rst
seanm: Whoops.

userguide/sharing.rst
seanm: I've reworked this whole section.

userguide/sharing.rst
seanm: OK, I copied the wording from the AFP section.

userguide/sharing.rst
seanm: I've reworked this whole section.

userguide/sharing.rst
seanm: Just because I did a find/replace of "Mac OS X" spelling doesn't mean I'm trying to find & correct every typo in this document...

userguide/sharing.rst
seanm: Will revert my changes to this paragraph.

userguide/sharing.rst
seanm: Again, I did a find/replace of "Mac OS X" spellings here, nothing more.  I'm a first time contributor to this project, minor style changes like that should be made by someone that knows the project's style rules...

wblock: Sorry, just wanted to be thorough.

userguide/sharing.rst
seanm: How does this sound to you:

* :ref:`AFP <Apple (AFP) Shares>`: Apple Filing Protocol shares are accessible from macOS computers (or from other operating systems with third party software). As of Mac OS X 10.9, Apple has deprecated AFP in favor of :ref:`SMB <Windows (SMB) Shares>`, and so use of AFP is no longer recommended unless 1) you need to support pre-10.9 clients or 2) you want to provide a Time Machine (since %brand% does not yet support Time Machine via SMB).

wblock: Again, this is a really long sentence.  We also try to avoid the informal "you" and asides.  So let me try some rewording:
```:ref:`AFP <Apple (AFP) Shares>`: Apple Filing Protocol shares are accessible from macOS computers or other operating systems with AFP software. Apple deprecated AFP at the release of Mac OS X 10.9 and recommends using :ref:`SMB <Windows (SMB) Shares>` instead. Use of AFP is not recommended unless pre-10.9 Mac clients must be supported or Apple's Time Machine, which requires AFP, is to be used.

(Note: I'm not sure about that last part. I thought Time Machine on SMB worked now.)

seanm: That wording is mostly fine with me, though your last sentence doesn't make clear that it's FreeNAS that forces AFP use, not Time Machine itself, how about:

"Use of AFP is not recommended unless pre-10.9 Mac clients must be supported or to provide a Time Machine, which %brand% currently only supports via AFP."

As best as I can tell SMB is not supported for Time Machine:
https://redmine.ixsystems.com/issues/23359

If you review the commit message, this lack of support is really the main reason for this patch. :)

I'm on the fence between buying a Synology or FreeNAS, mostly because of this.

Once FreeNAS supports Time Machine via SMB, there's no good reason to enable AFP, as it's only needed for OS X versions that don't even receive security updates from Apple anymore.

anodos325: Time machine over SMB is possible in 11.1. I tested it a few weeks ago. Backup completed successfully. It just requires some auxiliary parameters.

userguide/sharing.rst
anodos325: I believe that the whole heading "Configuring Authenticated Access Without a Domain Controller" is troublesome. What we're trying to show is "Configuring Authenticated Access with Local Users on the FreeNAS server". 

The current heading implies that if we have LDAP or NIS enabled, then we need to add local users as well (which we don't).

userguide/sharing.rst
seanm: Interesting. Where can I read about these 'some auxiliary parameters'?

Mrt134:
This PR has gone stale, so I'm porting the proposed changes and resolving the review in [PR 547](https://github.com/freenas/freenas-docs/pull/547). Thanks for the original submission seanm!

PR #146, opened 2018-06-19
Reviewed storage.rst:

sambonham:
Finally finished reviewing this massive section!
-Went through the entire document to verify the instructions match up with the UI, with the exception of the import geli-encrypted disk and view multi path sections. 
- Added too many of our new |ui-settings|, |ui-options|, |ui-options| icons.
- Removed "the following" references so Warren doesn't follow up with me. 
- Removed "you" and "you" references.
- Fixed incorrect use of possessives: disk's/dataset's/pool's.
- Replace "checkboxes" with "options".
- Replace "check" with "set" or "select" . 
- Fix minor typos like "iterms per page".
- Remove "it is recommended".
- Passed the build test with no errors.
- Checked for whitespaces with git dif --check

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -15,7 +15,7 @@ these options:
 
 * :ref:`Disks`: view and manage disk options.
 
-* :ref:`Importing a Disk`: import a **single** disk that has been
+* :ref:`Importing a Disk`: import a **single** disk that is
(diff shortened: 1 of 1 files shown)


dlavigne:
Don't forget to backport doc improvements to master. This one is a bit trickier as the order has changed in angulargui (and won't change in master), so just word improvements only please.

PR #165, opened 2018-06-22
Update Edit to reflect updated certificate fields

miker:
Certificate Identifiers are the only fields that are editable
Certificate and Private Key fields are read only


dlavigne:
Same comment as 166, remove the (name)

miker:
done and pushed

PR #166, opened 2018-06-22
Update system with current editable field

miker:
Only the Identifier field is editable for a Certificate


dlavigne:
I don't know when (name) snuck in, but it is not needed and is not part of the name of that field. Please shorten to:

certificate or to edit the :guilabel:`Identifier`.

miker:
It was in the original docs so I kept it. removed and pushed

PR #176, opened 2018-06-22
fix github link

miker:


miker:
wasn't sure if one creates a repo **on** or **at** github. Left the existing text using **on**.

PR #205, opened 2018-07-12
Update install.rst to be version agnostic in the "dd" examples:

Mrt134:

### DIFF
--- a/userguide/install.rst
+++ b/userguide/install.rst
@@ -103,10 +103,7 @@ On FreeBSD or Linux
 ~~~~~~~~~~~~~~~~~~~
 
 On a FreeBSD or Linux system, the :command:`dd` command is used to
-write the :file:`.iso` file to an inserted USB thumb drive. This
-example demonstrates writing the image to the first USB device
-(*/dev/da0*) connected to a FreeBSD system. Substitute the filename of
-the :file:`.iso` file and the device name of the device to write.
+write the :file:`.iso` file to an inserted USB thumb drive.
 
 .. warning:: The :command:`dd` command is very powerful and can
    destroy any existing data on the specified device. Make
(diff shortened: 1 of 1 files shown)


userguide/install.rst
```diff
@@ -115,9 +112,16 @@ the :file:`.iso` file and the device name of the device to write.
    command can be avoided by writing the :file:`.iso` file to a
    CD instead.
 
+
+This example demonstrates writing the image to the first USB device
+(*/dev/da0*) connected to a FreeBSD system. Substitute
+:literal:`FreeNAS-RELEASE.iso` with the filename of the downloaded
+:file:`.iso` file and :literal:`/dev/da0` with the device name of the
```
wblock: It's probably better to split the sentence here.

...file. Replace :literal:`/dev/da0` with ...

userguide/install.rst
```diff
@@ -206,26 +210,28 @@ and use the :command:`dd` command to write the image to the USB stick.
 In this example, the USB thumb drive is :file:`/dev/disk1`. It is
 first unmounted. The :command:`dd` command is used to write the
 image to the faster "raw" version of the device (note the extra
-:literal:`r` in :file:`/dev/rdisk1`). When running these commands,
-substitute the name of the installation file and the correct path to
-the USB thumb drive.
+:literal:`r` in :file:`/dev/rdisk1`).
+
+When running these commands, substitute :literal:`FreeNAS-RELEASE.iso`
```
wblock: s/substitute/replace/

Use the same language as above.

userguide/install.rst
```diff
@@ -206,26 +210,28 @@ and use the :command:`dd` command to write the image to the USB stick.
...
 
 .. note:: If the error "Resource busy" is shown when the
    :command:`dd` command is run, go to
    :menuselection:`Applications --> Utilities --> Disk Utility`,
    find the USB thumb drive, and click on its partitions to make sure
-   all of them are unmounted. If the error
-   "dd: /dev/disk1: Permission denied" is shown, run the :command:`dd`
-   command by typing
-   :command:`sudo dd if=FreeNAS-11.0-RELEASE.iso of=/dev/rdisk1 bs=64k`.
+   all of them are unmounted. If an error similar to
```
wblock: "similar to" really leaves the reader unsure.  Similar in what way?

If a "Permission denied" error is shown, use :command:`sudo` for elevated rights:

userguide/install.rst
```diff
@@ -206,26 +210,28 @@ and use the :command:`dd` command to write the image to the USB stick.
...
+   "dd: /dev/disk1: Permission denied" is shown, add :command:`sudo` in
+   front of the entire :command:`dd` command:
+   :command:`sudo dd if=FreeNAS-RELEASE.iso of=/dev/rdisk1 bs=64k`.
```
wblock: Use :samp: instead of :command:, and wrap the filename parameter with {} so it will be shown as replaceable.  Do that above, also.

PR #223, opened 2018-07-24
Rework resource link to only include it in the VM Wizard Options table.

Mrt134:
HTML build test: no issues.


dlavigne:
Master PR please.

PR #234, opened 2018-07-26
Fix 34384 [angulargui]:

Mrt134:
- Add text to description of "Disable Physical Block Size Reporting" field to note that setting can avoid warning spam with ESXi.
- Add link to relevant blog post.
- Fix a couple empty space issues.
- HTML build test: no issues.
- Change will also be ported to master branch.


dlavigne:
Will also need a master PR.

PR #271, opened 2018-07-31
Add link that further explains ssh-keygen

a-stjohn:
Added a link as mentioned in [forum post](https://forums.freenas.org/index.php?threads/public-private-ssh-key-for-freenas-user.62390/#post-445165).


dlavigne:
Is there an associated master PR? Is this related to a ticket or just a general improvement?

PR #281, opened 2018-08-09
Added text to reflect setting swap size 0 is bad.

a-stjohn:
Fix for [ticket 39493](https://redmine.ixsystems.com/issues/39493).


dlavigne:
This will also need a master PR.

PR #290, opened 2018-08-14
Improved text. Added note about changing default shell.

a-stjohn:


dlavigne:
This one also needs a master PR.

PR #294, opened 2018-08-15
Deleted user base and group base from the table.

a-stjohn:


dlavigne:
This will also need an entry in intro.rst, ordered by screen order in the section that describes the fields that have been added/removed.

dlavigne:
Please also create a PR for master.

dlavigne:
Or a confirmation that this is new UI only. Thanks!

PR #297, opened 2018-08-15
Updated docs with regards to ticket 41172.

a-stjohn:
Deleted the entry from the table about the telemetry checkbox. Also updated the screenshot.


dlavigne:
This will also need an entry in intro.rst in the section that indicates the fields that have changed. Insert the bullet to follow the order that the applicable screen appears in the UI. 

Once this PR is approved, a similar one will be needed for angulargui.

PR #298, opened 2018-08-15
Add entry in intro and update note in network to reflect Network/Inte…

Mrt134:
…rfaces change:

- Changes do not affect legacy UI or the master branch handbook.
- HTML build testing: no issues.


userguide/intro.rst
```diff
@@ -232,6 +232,12 @@ These screen options have changed:
   fields have been added to
   :menuselection:`Tasks --> Cloud Sync Tasks --> Add Cloud Sync`.
 
+* The :guilabel:`NIC` and :guilabel:`Interface Name` fields in
+  :menuselection:`Network --> Interfaces --> Add Interface`
+  are preconfigured with the web interface NIC settings when configuring
+  the first interface. The UI also warns the user to configure the web
+  interface to prevent connection issues.
```
wblock: Try to avoid saying "UI".  But also the web interface warning is only shown if the user tries to edit another interface first.  So:
"...interface. A warning is shown if changes are made to an additional interface before configuring the web interface."

userguide/network.rst
```diff
@@ -158,13 +158,12 @@ shows which interfaces are manually configured and allows adding
...
 .. note:: Typically, the interface used to access the %brand%
-   administrative GUI is configured by DHCP. This interface does not
-   appear in this screen, even though it is already dynamically
-   configured and in use.
-
-   :ref:`Creating a Link Aggregation` that does **not** include the NIC
-   used to access the %brand% administrative GUI may require adding an
-   :guilabel:`Interfaces` entry for this interface with DHCP enabled.
+   administrative GUI is configured by DHCP and not displayed in the
+   :guilabel:`Interfaces` list. The GUI will prompt to add this
```
wblock: As above, we should avoid "UI" and "GUI" as terms the user might not know. s/will prompt/prompts/, so:
"When additional network interfaces are added or configured, a prompt warns that the interface used for the administrative web page should be configured first to avoid losing the connection."

PR #309, opened 2018-08-17
Ticket 39563 fix. Shutdown screen

a-stjohn:


dlavigne:
No mention of fade to black after 30 seconds?

wblock:
After a merging of the log out, restart, and shut down chapters, and addition of the power icon, this patch has been overtaken by events and is no longer needed.

PR #341, opened 2018-08-30
Add text describing new Team Drive ID field in cloud_cred_tab.

Mrt134:
HTML build test: no issues


dlavigne:
Do not merge this until after this field is added to the UI for RC1.

PR #354, opened 2018-09-14
Update text about Sharing > NFS > Path field.

Mrt134:


dlavigne:
This will also need a similar PR for master.

PR #363, opened 2018-09-21
Document new Hostname field for UPS service configuration.

Mrt134:
HTML build test: no issues.


dlavigne:
If this is a new field, it will also need an entry in intro.rst.

Mrt134:
Ok, updated table entry in services.rst and added an entry to intro.rst.

dlavigne:
This will also need a master PR.

PR #364, opened 2018-09-21
Adjust Add Mount point documentation to show creating a new Destinati…

Mrt134:
…on directory.

HTML build test: no issues.


dlavigne:
Can an existing directory not be used?

Mrt134:
Restested and confirmed using an existing directory is possible - adjusting PR.

PR #378, opened 2018-10-01
Pluralized as many of the 'account' instances I could find.

a-stjohn:


dlavigne:
Did you try this command in your local branch? It should catch all of the instances (note that there is a space before the asterisk):
grep Account *

userguide/account.rst
```diff
@@ -1,9 +1,9 @@
-.. _Account:
+.. _Accounts:
 
-Account
-=======
+Accounts
+========
 
-The Account Configuration section of the |web-ui| describes
+The Accounts Configuration section of the |web-ui| describes
```
wblock: This is kind of weird.  It's just an Accounts section, "configuration" is not in there and should not be capitalized even if it was.

Also, the Accounts section of the webui does not "describe" how to manually manage users and groups, it just does it.

Suggested:
`` :menuselection:`Accounts` is used to manage users and groups. ``

wblock:
> Did you try this command in your local branch? It should catch all of the instances (note that there is a space before the asterisk):
> grep Account *

We also have more files in snippets/, so:

grep -r Account *

PR #399, opened 2018-10-12
Update Directory Service/LDAP/Hostname field text:

Mrt134:
HTML build test: no issues.


dlavigne:
While this is wrong, we need clarification on the correct text to use.

PR #406, opened 2018-10-22
Tried to clarify docs in regard to ticket 28649.

a-stjohn:


userguide/system.rst
```diff
@@ -417,9 +417,11 @@ configuration buttons are shown:
 
 The buttons above the boot entries can be used to:
 
-* **Create:** a manual boot environment. A pop-up menu prompts for
-  entry of a :guilabel:`Name` for the boot environment. Only
-  alphanumeric characters, underscores, and dashes are allowed.
+* **Create:** creates a copy of the active environment. The active boot
+  environment contains the text "On Reboot, Now" in the
```
wblock: Use the literal function for actual text: ` :literal:`On Reboot, Now` `

userguide/system.rst
```diff
@@ -432,6 +434,9 @@ The buttons above the boot entries can be used to:
   :numref:`Figure %s <status_boot_dev_fig>`,
   shows only one boot device, which is *ONLINE*.
 
+.. note:: Using :guilabel:`Clone` to clone the active boot environment
```
wblock: This is not exactly correct.  If it functions the same, we would not need two separate buttons.

Please expand on this.  I suspect it asks for a name, then clones a boot environment (only the active one?).

a-stjohn: Using "Clone" snaps the selected environment. (this includes selecting and cloning the active environment)
Using "Create" snaps the current "Active" environment.

a-stjohn: UPDATE: I've talked to Caleb about the functionality and to clarify things about it.
Essentially The "Create" and "Clone" button DO the same thing. However, the only difference is that if a user uses "Create" it makes a new boot environment based on the "Active" environment. The "Clone" button makes a new boot environment based on the boot environment selected.

I will put in a ticket.

PR #410, opened 2018-10-25
The enable power saving daemon checkbox was removed from the UI 

sambonham:
- I removed the enable power saving option from the Advanced Configuration Settings table. 
- Passed build tests with no errors.
- This is in response to Bug #49398


dlavigne:
Please add an entry in the intro to this PR.

PR #416, opened 2018-10-29
49851 docs changes:

Mrt134:
- Update vms-add-rancher.png
- Small updates to text in vms.rst.
- HTML build test: no issues.


dlavigne:
That picture does not match the label in the UI commit:
  label=_("Size of Docker Disk File (GiB)"),

Mrt134:
Fixed!

PR #424, opened 2018-10-29
Updated the docs to warn users against using uppercase usernames when…

sambonham:
… creating smb home shares.
Passed build test with no issues.
Related to Bug #45387


dlavigne:
Sam will also backport to master version of Guide.

PR #447, opened 2018-11-01
Port changes from PR 446 to angulargui branch:

Mrt134:
HTML build test: no issues.


dlavigne:
This will need to be reworked once the ticket passes testing. Additionally, the output in the freenas-debug section will need to be updated.

PR #453, opened 2018-11-06
Fix undefined label build error.

sambonham:
Passed build test with no errors.


dlavigne:
Adding Tim and Warren as reviewers as this fixes a build error.

dlavigne:
Tim has since fixed this on his latest PR.

PR #457, opened 2018-11-07
Add text about autorestore of iocage defaults for a jail with a corrupted config.

Mrt134:
HTMl build test: no issues.


userguide/jails.rst
```diff
@@ -611,6 +611,10 @@ The final set of jail properties are contained in the
 
 
 Click :guilabel:`SAVE` when the desired jail properties have been set.
+These properties are saved as a unique jail configuration. If an iocage
+jail configuration becomes corrupted, %brand% automatically restores
```
skarekrow: No. This is for the defaults file during creation. No bearing on an individual jail that has corruption.  (Yet)

PR #462, opened 2018-11-12
Remove text about "SAVE CONFIG" button and rework Manual Update secti…

Mrt134:
…on text.


dlavigne:
Hmm, what happened to the Save Config? Was it removed alltogether?

Mrt134:
Ok, per our conversation, I reworked the PR a bit to still mention the save config option in the Manual Update section.

PR #464, opened 2018-11-13
11.2 rc2 manual screenshots angular

Mrt134:
Review 50/50 manual screenshots and update 4 manual shots for RC2.
Add icon replacements for the new UI alert icons.
Remove 3 .png screenshots that are now unnecessary.
HTML and PDF build testing: no issues.


dlavigne:
Notes that we don't need the installer images as, IIRC, we already have ones for RELEASE which is right on the heels of RC2.

Mrt134:
Ah, ok - I'll change back installer-install-menu then.

PR #465, opened 2018-11-14
ES60 BSG Changes: #57150

Mrt134:
Address 3/4 change requests from Redmine Issue #57150:
- Add note to beginning of document that ES60 replaced the E60 and add the EOL and EOS dates for the E60.
- Reword paragraph about installing cage nuts to attempt to clarify. Investigation shows no photos available showing the specific process to install a cage nut.
- Rework two included snippets to fix erroneously visible ..index:: entries and avoid automatic hyphenation.
Still to do: Investigate and update description about pushing the unit inwards into the rack. This might require a significant rework to one of the more complicated photos or a reshoot of the image altogether.


userguide/snippets/es60.rst
```diff
@@ -126,8 +131,8 @@ secures the rail in place. Repeat this process on the other side.
 Mount Rack Rails
 ^^^^^^^^^^^^^^^^
 
-Install four cage nuts in the rack, one for each bracket. The lock
-screw for the bracket is already installed in the cage nut. Place the
+Install the four cage nuts with locking screws in the rack. Place one
```
wblock: "the" is not needed here, and might even confuse the reader if they get more than four nuts in a package.  The phrasing on the last part of the sentence is confusing also, implying there is a "locking screw", rather than a screw that holds the unit in place.

userguide/snippets/es60.rst
```diff
@@ -126,8 +131,8 @@ secures the rail in place. Repeat this process on the other side.
...
+cage nut for each bracket, where the rail will attach. Place the
```
wblock: The comma is really awkward here.  The rails can have four holes, so we should be specific about which one is used.  With those changes, we should be able to avoid future tense.

Mrt134:
Waiting for clarification about the rack rail mounting method before the "Mount Rack Rails" section can be fixed and the PR merged.

PR #474, opened 2018-11-21
Fix incorrect labels.

a-stjohn:


dlavigne:
This needs an angulargui equivalent.

PR #476, opened 2018-11-22
Split a paragraph for ease of reading

stilez:
This paragraph isnt easy to parse for a user.  Also it tends to mis-suggest key handling (it says it stores the key, but storing the key locally is actually a default+user choice, not a necessity, so the reader has to figure out this vs. next paragraph which is avoidable).

Refactor it (without changing its substance) to make it easier to read.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1371,17 +1371,21 @@ changed, and destroying a zvol requires confirmation.
 Managing Encrypted Volumes
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-%brand% generates and stores a randomized *encryption key* whenever
+%brand% generates a randomized *encryption key* whenever
 a new encrypted volume is created. This key is required to read and
-decrypt any data on the volume.
+decrypt any data on the volume. 
 
-Encryption keys can also be downloaded as a safety measure, to allow
-decryption on a different system in the event of failure, or to allow
-the locally stored key to be deleted for extra security. Encryption
-keys can also be optionally protected with a *passphrase* for
-additional security. The combination of encryption key location and
-whether a passphrase is used provide several different security
-scenarios:
+By default, %brand% stores encryption keys locally within its system
+settings, so that authorized users can easily login and decrypt volumes.
+
+Encryption keys can also be downloaded. This is useful as a safety
+measure, to allow key backup, decryption on a different system in the
+event of failure, or so that the locally stored key can be deleted for
+extra security. Encryption keys can also be optionally protected with a
+*passphrase* for additional security. 
+
+The combination of encryption key location and optional passphrase is
+useful in several different security scenarios:
 
 * *Key stored locally, no passphrase*: the encrypted volume is
   decrypted and accessible when the system running. Protects "data at
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1375,9 +1375,9 @@ Managing Encrypted Volumes
 a new encrypted volume is created. This key is required to read and
 decrypt any data on the volume. 
 
-By default, %brand% stores encryption keys locally within the system
-settings so volumes can automatically decrypt when the system
-starts or when triggered by an authorized user.
+%brand% stores encryption keys on the system boot pool by default, so 
+encrypted pools can be automatically decrypted when the system starts,
+or when triggered by an authorized user.
 
 Encryption keys can also be downloaded. This is useful as a safety
 measure to allow key backup, to use the key for decryption on a different system in the
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1376,7 +1376,7 @@ a new encrypted volume is created. This key is required to read and
 decrypt any data on the volume. 
 
 By default, %brand% stores encryption keys locally within the system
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
-Encryption keys can also be downloaded as a safety measure, to allow
-decryption on a different system in the event of failure, or to allow
-the locally stored key to be deleted for extra security. Encryption
-keys can also be optionally protected with a *passphrase* for
-additional security. The combination of encryption key location and
-whether a passphrase is used provide several different security
-scenarios:
+decrypt any data on the volume. 
+
+By default, %brand% stores encryption keys locally within its system
```
Mrt134:
```suggestion
%brand% stores encryption keys locally within the system
```

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
+settings, so that volumes can easily be decrypted automatically on system
```
Mrt134:
```suggestion
 settings so volumes can automatically decrypt when the system
```

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
+start-up, or by authorized users.
```
Mrt134:
```suggestion
starts or when triggered by an authorized user.
```

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
+
+Encryption keys can also be downloaded. This is useful as a safety
+measure, to allow key backup, decryption on a different system in the
+event of failure, or so that the locally stored key can be deleted for
+extra security. Encryption keys can also be optionally protected with a
```
Mrt134:
```suggestion
extra security. Encryption keys can also be protected with a
```

userguide/storage.rst
stilez: It doesn't always store it (you can remove the local copy). Surely better to say "by default stored" in the first place, rather than somewhat mislead with information that has to later be qualified, when its that easy to give accurately in the first place?

userguide/storage.rst
stilez: Yes

userguide/storage.rst
stilez: "The system" is good no question tho

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
-decryption on a different system in the event of failure, or to allow
-the locally stored key to be deleted for extra security. Encryption
-keys can also be optionally protected with a *passphrase* for
-additional security. The combination of encryption key location and
-whether a passphrase is used provide several different security
-scenarios:
+decrypt any data on the volume. 
+
+By default, %brand% stores encryption keys locally within the system
+ settings so volumes can automatically decrypt when the system
```
a-stjohn:
```suggestion
settings so volumes can automatically decrypt when the system
```

Appears to be an extra space there.

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
-decryption on a different system in the event of failure, or to allow
-the locally stored key to be deleted for extra security. Encryption
-keys can also be optionally protected with a *passphrase* for
-additional security. The combination of encryption key location and
-whether a passphrase is used provide several different security
-scenarios:
+decrypt any data on the volume. 
+
+By default, %brand% stores encryption keys locally within the system
+settings so volumes can automatically decrypt when the system
```
wblock: "within the system settings" is not clear.  Does it mean "in the system configuration database"?

s/can automatically decrypt/can be automatically decrypted/

This would also be an opportunity to avoid starting a sentence with a halting aside by moving the "by default" to later:

`` %brand% stores encryption keys on the system boot pool so encrypted pools can be automatically decrypted when the system starts. ``

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
+starts or when triggered by an authorized user.
+
+Encryption keys can also be downloaded. This is useful as a safety
+measure, to allow key backup, decryption on a different system in the
```
```suggestion
measure to allow key backup, to use the key for decryption on a different system in the
```

userguide/storage.rst
```diff
@@ -1371,17 +1371,22 @@ changed, and destroying a zvol requires confirmation.
...
+event of failure, or so that the locally stored key can be deleted for
```
```suggestion
event of failure, or so the locally stored key can be deleted for
```

userguide/storage.rst
stilez: Done

wblock:
This PR has been overtaken by changes in the previous PR about encrypted disks.  After talking to support, it was felt that dealing with manual mounting of encrypted pools is a complex topic that cannot be done through the UI, so we cannot give a complete description of it here.  One suggestion was submitting a document on that to the forums, which could be added to the Resources section.

Thanks!

PR #477, opened 2018-11-24
Googlification Prevention

aidik:
files containing html comment `<!--prevent googlize -->` will be ignored and not altered by Googlize.


wblock:
I added a version of this in bc2cc15 that works similarly.  It uses a case-insensitive match and allows different types of whitespace in the comment.  Also, the comment swaps the words around: "googlize prevent".  This is to be consistent with the other googlize comments.

PR #479, opened 2018-11-27
54807: Update tasks.rst text about redirect standard output/error:

Mrt134:
HTML build test: no issues.


Mrt134:
Affected image tasks-cron-jobs-add.png is part of screenshot automation and will be updated later.

PR #490, opened 2018-11-30
54954: Add "Enable SMB1 support" field to services.rst

Mrt134:
TrueNAS specific for now, so entry is behind ifdef statements.
Update tn_cifs1b.png


dlavigne:
Why the sysctl? Isn't that what the checkbox does for you?

Mrt134:
I originally had a larger warning in the table, but during review it was noted that this warning box existed and I should link back to it to avoid redundancy and open a ticket to move it (https://redmine.ixsystems.com/issues/60486).

dlavigne:
Right, but why are we telling users to set a sysctl? Isn't that the point of the checkbox? You are linking to this warning from the checkbox which makes a user think that they need to do both.

dlavigne:
Let's remember to use the same wording (and remove the ifdef in master) when we repeat for 11.2-U1.

PR #491, opened 2018-11-30
Update M-Series BSG from review notes (except image changes):

Mrt134:
25% reduction to title page image
Remove redundant word in section 1
Move table of storage controller connectors to after the back panel image.
Move single storage controller warning to after the back panel image.
Fix index entry that was erroneously visible.
PDF build testing: no issues.


wblock:
Overtaken by a rework of the component layout images.  Elements of this work were integrated into that work, though.

PR #497, opened 2018-12-03
Add TrueNAS specific text for the  Network/Interfaces section of netw…

Mrt134:
…ork.rst:

- Add tn_interface.png image and separate from the freenas image with ifdef statements.
- Rework Interface Configuration Settings table to include the TrueNAS specific fields.
- TrueNAS and FreeNAS HTML build testing: no issues.


dlavigne:
Do not merge until after U6.2 is released.

PR #503, opened 2018-12-04
Updated text about adding spare to enctypted pool. Also added new sym…

a-stjohn:
…bol in docs.

HTLM build: no issues
PDF build: no issues


dlavigne:
This still needs a master PR.

PR #510, opened 2018-12-17
Add note about how plugins and jails are updated/upgraded

a-stjohn:


userguide/plugins.rst
```diff
@@ -159,6 +159,11 @@ When a newer version of a plugin becomes available in the official
 repository, update the plugin jail by navigating to the
 :menuselection:`Jails` page and clicking |ui-options| and
 :guilabel:`Update`.
+
+.. note:: Plugins are updated using :samp:`iocage update {UUID} | {NAME}`
+   and upgraded using :samp:`iocage upgrade {UUID} | {NAME} -r {RELEASE}`.
```
skarekrow: plugin upgrades don't supply a RELEASE. iocage upgrade PLUGIN`

userguide/plugins.rst
```diff
@@ -159,6 +159,11 @@ When a newer version of a plugin becomes available in the official
...
+   Manually created jails use :samp:`pkg update` and :samp:`pkg upgrade`.
```
skarekrow: They use the same iocage interfaces as plugins. iocage adapts in the background. So this line can be removed.

userguide/plugins.rst
```diff
@@ -159,6 +159,10 @@ When a newer version of a plugin becomes available in the official
 repository, update the plugin jail by navigating to the
 :menuselection:`Jails` page and clicking |ui-options| and
 :guilabel:`Update`.
+
+.. note:: Plugins are updated using :samp:`iocage update {PLUGIN}`
```
wblock: Wait... this is not clear on several things.  What is the difference between an "update" and an "upgrade"?  Is this aimed at the end user or the plugin developer?  Does the end user have to use the command line to update or upgrade plugins?

userguide/plugins.rst
```diff
@@ -160,8 +160,10 @@ repository, update the plugin jail by navigating to the
 :menuselection:`Jails` page and clicking |ui-options| and
 :guilabel:`Update`.
 
-.. note:: Plugins are updated using :samp:`iocage update {PLUGIN}`
-   and upgraded using :samp:`iocage upgrade {PLUGIN}`.
+.. note:: Plugins can also be updated using :samp:`iocage update {PLUGIN}`
```
wblock: I don't think this needs to be a note.  Instead, break it into two paragraphs: one that describes what an update is and how it is done, and a second that describes what an upgrade is and how to do it.

PR #515, opened 2018-12-18
Added the physical weight of each system to the warning about team lifting.

sambonham:
Also expanded on the warning message based on feedback from Andrew N.
These changes were based on a request made in https://redmine.ixsystems.com/issues/57330

### DIFF
--- a/userguide/snippets/es24m_rails.rst
+++ b/userguide/snippets/es24m_rails.rst
@@ -52,10 +52,15 @@ click in place. Repeat this process for the second rear rail.
 Mount the Unit in the Rack
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**Caution: Two people are required to safely lift the chassis for rack
-installation or removal.** Do not install drives until after the
-chassis has been installed in the rack, and remove all drives before
-removing the chassis from the rack.
+.. caution:: The unit weighs about 40 pounds.
+   At least two people are required to safely lift the
+   chassis for rack installation or removal. This process
+   should not be attempted alone as it involves risks and
+   could result in equipment damage or personal injury.
+
+Do not install drives until after the chassis has been installed
+in the rack, and remove all drives before removing
+the chassis from the rack.
 
 Pull the front rack rail forward until it stops. Align the cabinet
 rail with the inside of the front rack rail and slide the cabinet rail
@@ -81,4 +86,3 @@ pushing the unit in until it is flush with the front of the rack.
 
 Anchor the unit in the rack on both sides with the M5x18 screws
 included in the rail kit.
-
--- a/userguide/snippets/es24m_rails.rst
+++ b/userguide/snippets/es24m_rails.rst
@@ -52,11 +52,11 @@ click in place. Repeat this process for the second rear rail.
 Mount the Unit in the Rack
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. caution:: The unit weighs about 40 pounds.
-   At least two people are required to safely lift the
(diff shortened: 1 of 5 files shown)


userguide/snippets/es24m_rails.rst
```diff
@@ -52,10 +52,15 @@ click in place. Repeat this process for the second rear rail.
 Mount the Unit in the Rack
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**Caution: Two people are required to safely lift the chassis for rack
-installation or removal.** Do not install drives until after the
-chassis has been installed in the rack, and remove all drives before
-removing the chassis from the rack.
+.. caution:: The unit weighs about 40 pounds.
```
Mrt134: Spec sheet for ES24 says 40 lbs unloaded, 76 lbs when fully loaded. At minimum, I'd remove "about" -> "The unit weighs 40 pounds." Might want to rework a bit to show the loaded weight too: "The unit weighs 40 pounds empty and 76 pounds when fully loaded with drives."

userguide/snippets/es60.rst
```diff
@@ -139,10 +140,15 @@ rack holes until they lock in place.
 Mount Unit in the Rack
 ~~~~~~~~~~~~~~~~~~~~~~
 
-**Caution: Two people are required to safely lift the chassis for rack
-installation or removal.** Do not install drives until after the
-chassis has been installed in the rack, and remove all drives before
-removing the chassis from the rack.
+.. caution:: The unit weighs about 80 pounds.
```
Mrt134: As above, spec sheet states 85 lbs unloaded and 175 lbs loaded.

userguide/snippets/unified-storage-array.rst
```diff
@@ -168,9 +168,12 @@ in the rack and secure them with 8 #32 thumbscrews.
 
 The %brand% Storage Array can now be installed into the rack.
 
-.. caution:: Two people are required to safely install or remove the
-   %brand% Storage Array in a rack or cabinet.
-
+.. caution:: The unit weighs about 65 pounds.
```
Mrt134: As above, I don't think "about" is necessary.

userguide/snippets/x_rails.rst
```diff
@@ -23,25 +22,26 @@ Install a spring on the silver posts in the side of each rail.
...
 pins on both ends of the rail with the mounting holes in the rack.
 Swing the clamp latch closed to hold the rail in place. Use two of the
 supplied screws to secure the back end of the rail in place. Repeat
 the process for the second rail.
 
-
 .. figure:: images/tn_x_railclamp.png
    :width: 100%
 
+.. caution:: The unit weighs about 30 pounds.
```
Mrt134: Same as above, exact weight from the spec sheet says 26 lbs unloaded, 44 lbs loaded.

userguide/snippets/es24m_rails.rst
Mrt134: Reviewed M-Series BSG and noticed it uses this file for the rack mount instructions too. This caution needs to show weights for both the ES24 and the M-Series.

wblock: There is some margin for error and changes due to packaging.  The "about" is good to show that some variation is expected and the quoted weights should be verified before someone ships one.

wblock:
Nobody knows where the idea to put exact weights came from.  Due to potentially misleading the reader, and because we already have warnings that two people are required to rack a system, I'm closing this PR.  If it is resubmitted, we can follow up with the requester.

PR #520, opened 2018-12-19
Update Storage/Managing Encrypted Pools to better explain ZFS "full disk" encryption.

Mrt134:
Also provide a reference to the Self-Encrypting Drives section.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -209,9 +209,12 @@ a pool with encryption:
 * %brand% encryption is different from the encryption used in
   Oracle's proprietary, non-open source version of ZFS.
 
-* In %brand%, entire disks are encrypted, not individual filesystems.
-  Encrypted devices are created from the underlying drives, then the
-  pool is created on top of the encrypted devices.
+* In %brand%, entire disk partitions are encrypted, not individual
+  filesystems. Encrypted devices are created from the underlying drives,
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -209,9 +209,12 @@ a pool with encryption:
 * %brand% encryption is different from the encryption used in
   Oracle's proprietary, non-open source version of ZFS.
 
-* In %brand%, entire disks are encrypted, not individual filesystems.
-  Encrypted devices are created from the underlying drives, then the
-  pool is created on top of the encrypted devices.
+* In %brand%, entire disk partitions are encrypted, not individual
+  filesystems. Encrypted devices are created from the underlying drives,
+  then the pool is created on top of the encrypted devices. The drive is
+  not entirely encrypted as there will be a GPT partition used to
```
a-stjohn:
```suggestion
  not entirely encrypted as there is a GPT partition used to
```

wblock: Better for tense, but I think saying "the drive is not entirely encrypted" could lead the reader to think that not all of their data is encrypted.  Explain how it really is: a partition table is created on an unencrypted drive. Data partitions are then created, and encrypted data is stored in those partitions.  These are generally called "encrypted drives", even though the partition table is not encrypted.

PR #535, opened 2019-01-02
Clarified the cloud sync scheduling function.

a-stjohn:
Fixing ticket#29074


dlavigne:
This also needs a master equivalent.

PR #546, opened 2019-01-07
Added note about cloud sync operation times (master).

a-stjohn:


userguide/tasks.rst
```diff
@@ -168,7 +168,12 @@ shows the configuration options for Cloud Syncs.
    | Enabled                | checkbox            | Unset to temporarily disable this Cloud Sync.                                                          |
    |                        |                     |                                                                                                        |
    +------------------------+---------------------+--------------------------------------------------------------------------------------------------------+
-
+   
+   
+.. note:: The time chosen for the Cloud Sync refers to the time that
+   the task is able to start syncing. It does not mean that is when the
```
Mrt134:
```suggestion
   the task is able to start syncing. It is not when the
```

wblock: I suspect we would be better off without the second sentence at all.  We tell the reader what it is, so it is not necessary to tell them what it is not.

In the first sentence, it is clearer if s/is able to/is allowed to/

userguide/tasks.rst
```diff
@@ -168,7 +168,11 @@ shows the configuration options for Cloud Syncs.
    | Enabled                | checkbox            | Unset to temporarily disable this Cloud Sync.                                                          |
    |                        |                     |                                                                                                        |
    +------------------------+---------------------+--------------------------------------------------------------------------------------------------------+
-
+   
+   
+.. note:: The time chosen for the Cloud Sync refers to the time that
```
wblock: This is kind of redundant and not very clear.

`` The time selected is when the Cloud Sync task is allowed to begin. See :ref:`Cron Jobs` for setting the time when Cloud Syncs are attempted. ``

Please verify that this is correct (Cloud Sync times are set as cron jobs, not in Cloud Sync).

a-stjohn: I spoke with William. The Cloud Sync time chosen is when the task will run. It has nothing to do with Cron Jobs in the UI.

PR #553, opened 2019-01-08
Added a warning that credentials are lost when key isn't preserved (11.1-stable).

a-stjohn:


userguide/system.rst
```diff
@@ -1618,6 +1618,9 @@ The credentials to provide secure connections with cloud services
 are entered here. Amazon S3, Azure Blob Storage, Backblaze B2, and
 Google Cloud Storage are supported.
 
+.. warning:: Cloud credentials are encrypted. The credentials are lost
+   when the the encryption key is not preserved.
```
wblock: It's not clear what "preserved" means here.  Is this saying that if the encryption key is lost, the credentials will be lost also?

PR #555, opened 2019-01-08
Added warning that credentials are lost when key isn't preserved (master).

a-stjohn:


userguide/system.rst
```diff
@@ -1589,6 +1589,11 @@ Dropbox, FTP, Google Cloud Storage, Google Drive, HTTP, Hubic, Mega,
 Microsoft Azure Blob Storage, Microsoft OneDrive, pCloud, SFTP, WebDAV,
 and Yandex are supported.
 
+.. warning:: Cloud Credentials are stored in encrypted form. To be able
+   to restore Cloud Credentials from a
+   :ref:`saved configuration<General>`, "Export Password Secret Seed"
```
wblock: There should be a space between the label and the link, although it might work without that.
```suggestion
   :ref:`saved configuration <General>`, "Export Password Secret Seed"
```

a-stjohn:
Checked documentation and it said not to put a space there. Regardless, it
builds with no errors and operates as intended.

PR #556, opened 2019-01-08
Added warning about using manual volume creation (11.1-stable).

a-stjohn:


userguide/storage.rst
```diff
@@ -312,6 +312,10 @@ such volumes before using them in production.
 Manual Setup
 ^^^^^^^^^^^^
 
+.. warning:: Make sure that the disks are correctly setup by verifying
```
wblock: This is the same message as was proposed to be added as a warning to the code.  But it does not tell the user anything useful.  How do they know whether their selections are set up correctly?

(Also: "set up" with a space when it is a verb.)

wblock:
Overtaken by events.

PR #560, opened 2019-01-09
Add max. character limit to IPMI password field

Mrt134:
HTML build test: no issues.


dlavigne:
This will also need an 11.1-stable PR

Mrt134:
Just checked and this change is already present in freenas/11.1-stable branch.

PR #563, opened 2019-01-09
Removed warning about system dataset (11.1-stable).

a-stjohn:


dlavigne:
This only applies to TrueNAS.

PR #574, opened 2019-01-10
Add text about character restrictions for SMTP passwords.

Mrt134:
HTML build test: no issues.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -910,8 +910,8 @@ shown in
    | Username             | string               | Enter the SMTP username if the SMTP server requires authentication.                             |
    |                      |                      |                                                                                                 |
    +----------------------+----------------------+-------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


userguide/system.rst
```diff
@@ -910,7 +910,8 @@ shown in
    | Username             | string               | Enter the SMTP username if the SMTP server requires authentication.                             |
    |                      |                      |                                                                                                 |
    +----------------------+----------------------+-------------------------------------------------------------------------------------------------+
-   | Password             | string               | Enter the SMTP password if the SMTP server requires authentication.                             |
+   | Password             | string               | Enter the SMTP password if the SMTP server requires authentication. Accepts 8-bit ASCII symbols |
+   |                      |                      | (:literal:`!,",#,...`) and alphanumeric (:literal:`Abc123...`) plain text characters only.      |
```
wblock: It is misleading to list a few characters out of the bunch, and impractical to list all of them.  The error message from the source is clearer:
`` Only plain text characters (8-bit ASCII) are allowed in passwords. UTF or composed characters are not allowed. ``

PR #578, opened 2019-01-14
56610: Include snippet about User Guide location in "Support Resources" chap…

Mrt134:
…ter.

- Rework snippet to include an ifdef for the TrueNAS guide link.
- FreeNAS HTML build test: no issues.
- Master branch already has an equivalent, no port to that branch is necessary.

### DIFF
--- a/userguide/snippets/userguide.rst
+++ b/userguide/snippets/userguide.rst
@@ -9,7 +9,9 @@ User Guide
 The %brand% User Guide with complete configuration instructions is
 available either by clicking :guilabel:`Guide` in the %brand% user
 interface or going to
+#ifdef freenas
 `<https://www.ixsystems.com/documentation/freenas/>`__
+#endif freenas
 #ifdef truenas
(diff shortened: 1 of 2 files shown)


userguide/snippets/userguide.rst
```diff
@@ -9,4 +9,7 @@ User Guide
 The %brand% User Guide with complete configuration instructions is
 available either by clicking :guilabel:`Guide` in the %brand% user
 interface or going to
+`<https://www.ixsystems.com/documentation/freenas/>`__
+#ifdef truenas
 `<https://support.ixsystems.com/truenasguide/truenas.html>`__.
```
wblock: This needs to be two #ifdefs, one for FreeNAS and one for TrueNAS.
```
#ifdef freenas
`<https://www.ixsystems.com/documentation/freenas/>`__
#endif freenas
#ifdef truenas
`<https://www.ixsystems.com/documentation/truenas/>`__
#endif truenas

PR #619, opened 2019-01-22
Made 'Alert Settings' into refs.

a-stjohn:


userguide/snippets/alertevents.rst
```diff
@@ -81,7 +81,7 @@ Some of the conditions that trigger an alert include:
   goes red at 95%
 
 * new OpenZFS feature flags are available for the pool; this alert can
-  be adjusted in :guilabel:`Alert Settings` if a pool upgrade is not
+  be adjusted in :ref:`Alert Settings<Alert Settings>` if a pool upgrade is not
```
wblock: As above:
```suggestion
  be adjusted in :ref:`Alert Settings` if a pool upgrade is not
```

PR #631, opened 2019-01-23
Add entry reflecting l2arc size reporting change to intro.

Mrt134:
Investigation of storage.rst and reporting.rst shows no changes needed.
Investigation of angulargui branch and 11.1-stable branch shows change not reflected in either FreeNAS version.
HTML build test: no issues.


dlavigne:
This needs an angulargui equivalent.

PR #646, opened 2019-01-28
63483: Document new S3 Credential checkboxes (master)

Mrt134:
- Add entries to Cloud Credential Options table in system.rst
- Add entry to intro.rst U2 section.
- HTML build test: no issues.

### DIFF
--- a/userguide/intro.rst
+++ b/userguide/intro.rst
@@ -357,7 +357,7 @@ U2
 * The :guilabel:`Endpoint does not support regions` and
   :guilabel:`Use v2 signatures` fields have been added to
   :menuselection:`System --> Cloud Credentials --> Add Cloud Credential`
-  when the :guilabel:`Provider` is *Amazon S3*.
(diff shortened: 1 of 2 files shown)


userguide/intro.rst
```diff
@@ -357,7 +357,7 @@ U2
 * The :guilabel:`Endpoint does not support regions` and
```
wblock: I thought we changed that, please verify field label is correct here.

userguide/intro.rst
```diff
@@ -357,7 +357,7 @@ U2
...
   :guilabel:`Use v2 signatures` fields have been added to
```
wblock: Please verify field label text and capitalization.

userguide/intro.rst
Mrt134: Rechecked: it is correct for the legacy UI

userguide/intro.rst
Mrt134: Rechecked: it is correct for the legacy UI

PR #652, opened 2019-01-29
53638: Add entry to intro to note pool widgets in dashboard change colors

Mrt134:
HTML build test: no issues.

Specific change to Angular GUI; no port to master branch needed.

### DIFF
--- a/userguide/intro.rst
+++ b/userguide/intro.rst
@@ -410,7 +410,7 @@ U2
   :menuselection:`Reporting`
   now shows the compressed physical L2ARC size.
 
-* The pool widgets in the
+* Pool widgets in the
   :menuselection:`Dashboard`
   now change color to reflect the current pool status.
 
--- a/userguide/intro.rst
+++ b/userguide/intro.rst
@@ -410,6 +410,10 @@ U2
   :menuselection:`Reporting`
   now shows the compressed physical L2ARC size.
 
+* The pool widgets in the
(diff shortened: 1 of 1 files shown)


userguide/intro.rst
```diff
@@ -410,6 +410,10 @@ U2
   :menuselection:`Reporting`
   now shows the compressed physical L2ARC size.
 
+* The pool widgets in the
```
```suggestion
* Pool widgets in the
```

dlavigne:
No mention at all in the Storage chapter? Inquiring minds will wonder which color represents what (which is specified in the acceptance criteria for this ticket).

wblock:
We can't mention specific colors, because colors are themed and none of them should be hard-coded due to allowing for color-blind users.  Someone using a black-and-white high contrast theme will not see the same colors as with other themes, and every theme can have different emphasis or "problem" colors.

dszidi:
A couple of things to note...
- The Storage pool widgets specifically have colors changing based on status (red, yellow & green) so I think it would be worth mentioning
- It is entirely possible to put in some overrides to make the high contrast theme have some hard coded values. To accommodate color blind users, those alternate colors can be mentioned if need be

Mrt134:
I was hesitant to put in the specific colors because they are pretty different between each theme, but I'll add some text to the storage chapter to try and explain it.

Mrt134:
https://github.com/freenas/freenas-docs/pull/656

PR #666, opened 2019-01-31
50321: Add text about new NAA column in Sharing/iSCSI/Extents:

Mrt134:
- Add entry in intro.rst/U2 section.
- Add a bit of text at the end of the sharing.rst/Block (iSCSI)/Extents section
  explaining where the NAA is shown.
- HTML build test: no issues


Mrt134:
This change is not reflected in the legacy UI

PR #675, opened 2019-02-01
Doc Queue item 77: FreeNAS forum: Clarify initial SED password

Mrt134:
(Cherry-pick PR674 to master)
- Adjust terminology to reflect that the SED is unlocked until the administrator "takes ownership" by setting the password.
- This should avoid implying the device is modifying/removing existing data.
- HTML build test: no issues.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -752,7 +752,9 @@ to access the full capabilities of the device. %brand% provides the
 the command line.
 
 By default, SED devices are not locked until the administrator takes
-ownership by explicitly configuring a global or per-device password.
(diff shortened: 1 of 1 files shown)


userguide/system.rst
```diff
@@ -804,7 +806,7 @@ column of :menuselection:`Storage --> View Disks`. Conversely, the rows
 in that column will be empty for disks that do not support SED or which
 are unlocked using the global password.
 
-Next, remember to initialize the devices:
+Next, remember to take ownership of the devices:
```
```suggestion
Remember to take ownership of the devices:
```

PR #676, opened 2019-02-03
Small edits to Virtual Machines page

hartzell:
1. While the docs say that the docker vm needs to be "stopped" to change the password, I needed to actually needed to select 'Power Off' to get access to the Devices page.

2. Docs said "radial button" when they almost certainly meant "radio" button.


Mrt134:
An earlier PR reviewed this file and appears to have added the proposed changes. We're going to close this PR, but thank you for the submission!

PR #692, opened 2019-02-06
Update text on editing running jails. Read-only when running.

a-stjohn:


userguide/jails.rst
```diff
@@ -737,9 +737,9 @@ available for a jail.
    |              |                                                               |
    +==============+===============================================================+
    | Edit         | Used to modify the settings described in                      |
-   |              | :numref:`Table %s <jail_overview_tab>`. A popup error will    |
-   |              | display if the jail's :guilabel:`Status` is *up*.             |
-   |              | A jail cannot be edited while it is running.                  |
+   |              | :numref:`Table %s <jail_overview_tab>`.                       |
+   |              | A jail cannot be edited while it is running, but the settings |
```
```suggestion
   |              | A jail cannot be edited while it is running. The settings can |
```

userguide/jails.rst
```diff
@@ -737,9 +737,9 @@ available for a jail.
...
+   |              | can be viewed as read only.                                   |
```
```suggestion
   |              | can be viewed, but are read-only.                             |
```

PR #697, opened 2019-02-08
Update all needed manual screenshots for 11.2-U2.

Mrt134:
Worked off Angular GUI Master Image list.


Mrt134:
Saw an error in one image. Fixing it now.

PR #703, opened 2019-02-08
Update the old webui screenshot to exclude the vcenter plugin.

a-stjohn:


Mrt134:
If it matters, the new shot has the yellow "Warning" light in the upper-right. Everything else looks ok.

wblock:
Font and overall screenshot size are smaller.  We should not have examples that show a yellow alert, either.

PR #730, opened 2019-02-27
27864 (TrueNAS 11.2-U3): Add instruction to use physical interface IP addressses only w…

Mrt134:
…ith iSCSI ALUA.

TrueNAS guide build testing: no issues.

### DIFF
--- a/userguide/sharing.rst
+++ b/userguide/sharing.rst
@@ -1781,8 +1781,8 @@ link :guilabel:`Add extra Portal IP`.
    |                | menu         | address of *0.0.0.0* (any interface).                                           |
    |                |              |                                                                                 |
 #ifdef truenas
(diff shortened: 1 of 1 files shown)


userguide/sharing.rst
```diff
@@ -1780,6 +1780,10 @@ link :guilabel:`Add extra Portal IP`.
    | IP address     | drop-down    | Select the IPv4 or IPv6 address associated with an interface or the wildcard    |
    |                | menu         | address of *0.0.0.0* (any interface).                                           |
    |                |              |                                                                                 |
+#ifdef truenas
+   |                |              | Choose only physical interface IP addresses when configuring iSCSI ALUA.        |
```
wblock: We should mention not to use the VIP address specifically.

PR #736, opened 2019-03-05
73650: Add new entry for "Access Based Share Enumeration" checkbox in…

Mrt134:
... the SMB Share Options table.

HTML build test: no issues.


dlavigne:
This will also need an entry in the U3 section of the intro.

userguide/sharing.rst
```diff
@@ -928,6 +928,10 @@ provides more details for each configurable option.
    | Only Allow Guest Access        | checkbox      | ✓        | Requires :guilabel:`Allow guest access` to also be enabled. Forces guest access for all connections.                                                 |
    |                                |               |          |                                                                                                                                                      |
    +--------------------------------+---------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------+
+   | Access Based Share Enumeration | checkbox      | ✓        | Restrict share visibility to Windows users with read or write access to the share, based on the current Windows Share ACL. Use Windows               |
```
wblock: The last clause makes it a bit unclear. Does "Restrict share visibility to Windows users with current Windows Share ACL access of read or write." capture the correct meaning of this?

Also, not sure if Share should be capitalized in this or the next sentence.

PR #737, opened 2019-03-05
73650: Update "Access Based Share Enumeration" field description

Mrt134:
Synchronize text between legacy and new UIs.
HTML build test: no issues.

### DIFF
--- a/userguide/sharing.rst
+++ b/userguide/sharing.rst
@@ -960,8 +960,8 @@ provides more details for each configurable option.
    | Only Allow              | checkbox     | ✓           | Requires :guilabel:`Allow guest access` to also be enabled. Forces guest access for all connections.                                              |
    | Guest Access            |              |             |                                                                                                                                                   |
    +-------------------------+--------------+-------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


userguide/sharing.rst
```diff
@@ -960,8 +960,8 @@ provides more details for each configurable option.
...
    | Guest Access            |              |             |                                                                                                                                                   |
    +-------------------------+--------------+-------------+---------------------------------------------------------------------------------------------------------------------------------------------------+
-   | Access Based            | checkbox     | ✓           | When enabled, users can only see the shares they have permission to access. To change the default that grants everyone access, use the computer   |
-   | Share Enumeration       |              |             | management MMC on Windows or the :command:`sharesec` command-line utility.                                                                        |
+   | Access Based            | checkbox     | ✓           | Restrict share visibility to Windows users with read or write access to the share, based on the current Windows Share ACL. Use Windows            |
```
seanm: Might I suggest removing the first "Windows" there. i.e. "Restrict share visibility to users with read or write..."

This option applies to macOS clients too (I've verified). Presumably, it applies to all clients, regardless of their OS.

Mrt134: I've discussed with Engineering and confirmed the first part of the text can be Windows agnostic. Linking PRs here.

PR #743, opened 2019-03-06
Updated sftp field requirement wording. It changed from an 'and' to 'or'

a-stjohn:


dlavigne:
There is a spurious period at the end of the second column in the SFTP entry.

PR #747, opened 2019-03-07
69186: Update LUN ID field description

Mrt134:
Description in Associated Target Configuration Settings table now matches UI help text.
HTML build test: no issues.


userguide/sharing.rst
```diff
@@ -2171,9 +2171,8 @@ and extents.
    | Target      | drop-down menu | Select an existing target.                             |
    |             |                |                                                        |
    +-------------+----------------+--------------------------------------------------------+
-   | LUN ID      | integer        | Type a value between *0* and *1023*. Note that some    |
-   |             |                | initiators expect a a value below *256*. Enter *0*     |
-   |             |                | to statically assign the next available ID.            |
+   | LUN ID      | integer        | Select the value or enter a value between 0 and 1023.  |
```
wblock: "value" is redundant.  How about just "Select or enter a value..."?

userguide/sharing.rst
```diff
@@ -2171,9 +2171,8 @@ and extents.
...
+   |             |                | Some initiators expect a value below 256.              |
```
wblock: s/below/less than/

PR #749, opened 2019-03-07
62874: Create 'Network Summary' section.

a-stjohn:


userguide/network.rst
```diff
@@ -737,6 +737,18 @@ More information about this command can be found at
...
 
+.. _Network Summary:
+
+Network Summary
+---------------
+
+:menuselection:`Network --> Network Summary`
+shows a quick summary of the addressing information of every
+configured interface. For each interface name, the configured IPv4 and
+IPv6 addresses, DNS servers, and default gateway are displayed. It
```
Mrt134: Is "default gateway" supposed to be the "Default Routes" I see in the page, or is "Default Routes" another object?

userguide/network.rst
```diff
@@ -737,6 +737,18 @@ More information about this command can be found at
...
+also displays nameservers.
```
wblock: This last sentence doesn't make sense.  DNS servers *are* nameservers.  Should also put DNS default gateway with the other network information:
"IPv6 addresses, default gateway, and DNS nameservers."

userguide/network.rst
wblock: Default gateway and default router are the same thing.  Default routes are actual routes, but I don't think that term was used.  IP addresses and default gateway are basic information, DNS nameservers are secondary.

PR #750, opened 2019-03-07
65214: Add text for refresh button on updates page.

a-stjohn:


dlavigne:
This one also needs a mention in intro as it is a new button.

PR #751, opened 2019-03-07
Added short section on zdb to Command Line Utilities under freenas-debug with correct cache location flag

ZackaryWelch:


userguide/cli.rst
```diff
@@ -985,6 +985,17 @@ To collect the output of every module, use :literal:`-A`:
 
    freenas-debug -A
 
+For collecting debug information about a single pool, use
+:command:`zdb` with :literal:`-U /data/zfs/zpool.cache`
+followed by the name of the pool:
```
wblock: In the old UI, we are still using the term "volume".  The change to this branch should mention both:
"followed by the name of the volume (ZFS pool):"

userguide/cli.rst
```diff
@@ -985,6 +985,17 @@ To collect the output of every module, use :literal:`-A`:
...
+
+.. code-block:: none
+
+    zdb -U /data/zfs/zpool.cache pool1
```
wblock: In master, I think the example volume name is "volume1". Please check.

ZackaryWelch: Actually pool1 is used much more often than volume1 apparently

ZackaryWelch: Changed it to volume1 anyway.

PR #757, opened 2019-03-07
58437: Add entry for new CPU temperature graphs.

Mrt134:
Also make small change in reporting to mention the new graphs.
HTML build test: no issues.

### DIFF
--- a/userguide/reporting.rst
+++ b/userguide/reporting.rst
@@ -32,8 +32,8 @@ Graphs are grouped by category on the Reporting page:
   * `CPU <https://collectd.org/wiki/index.php/Plugin:CPU>`__
     shows the amount of time spent by the CPU in various states
     such as executing user code, executing system code, and being
-    idle. There are also CPU load graphs for short, mid, and long terms
-    and CPU temperature graphs.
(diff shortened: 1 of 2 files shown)


userguide/reporting.rst
```diff
@@ -32,8 +32,8 @@ Graphs are grouped by category on the Reporting page:
   * `CPU <https://collectd.org/wiki/index.php/Plugin:CPU>`__
     shows the amount of time spent by the CPU in various states
     such as executing user code, executing system code, and being
-    idle. It also shows the load on the CPU for a short term, mid term,
-    and long term.
+    idle. There are also CPU load graphs for short, mid, and long terms
```
wblock: These should be hyphenated, and the "There are" is a little awkward.
```suggestion
    idle. Graphs of short-, mid-, and long-term load are shown, along with
```

userguide/reporting.rst
```diff
@@ -32,8 +32,8 @@ Graphs are grouped by category on the Reporting page:
...
+    and CPU temperature graphs.
```
wblock:
```suggestion
    CPU temperature graphs.
```

PR #761, opened 2019-03-08
74938: Add warning about trying to configure Domain Controller when A…

Mrt134:
…D Monitoring is set.

HTML build test: no issues.

Functionality is not present in legacy UI. No master branch PR is necessary.


dlavigne:
This needs a master equivalent.

PR #762, opened 2019-03-08
76701: Add Restart option to a Running VM.

Mrt134:
HTML build test: no issues.

### DIFF
--- a/userguide/virtualmachines.rst
+++ b/userguide/virtualmachines.rst
@@ -81,6 +81,8 @@ are available:
 
 * :guilabel:`Stop` shuts down the VM.
 
+* :guilabel:`Restart` shuts down and immediately starts the VM.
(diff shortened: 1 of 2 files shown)


userguide/virtualmachines.rst
```diff
@@ -81,6 +81,8 @@ are available:
 
 * :guilabel:`Stop` shuts down the VM.
 
+* :guilabel:`Restart` shuts down and immediately starts the VM.
```
a-stjohn: I added a description for this option as well in PR #765. Be aware that other PR's prob conflict.

Mrt134: Let's keep this PR open and just back that change out of #765 . That will prevent having to update all the Redmine tickets and doc queue.

PR #776, opened 2019-03-13
Updated text and screenshots for the ESXi section.

a-stjohn:


userguide/install.rst
```diff
@@ -999,115 +999,91 @@ option. You can permanently boot from disk by removing the
...
-
-
-.. _esxi_new_vm_fig:
-
-.. figure:: images/vmware-configuration.png
-
-   New Virtual Machine Wizard
+<https://www.vmware.com/products/esxi-and-esx.html>`__.
 
+When the VMware vSphere client is installed, use it to connect to the
```
wblock: Can we reword this to avoid the if/then (when)?  Just "Install and use th VMware vSphere client to connect..."

userguide/install.rst
```diff
@@ -999,115 +999,91 @@ option. You can permanently boot from disk by removing the
...
+ESXi server. Enter the username and password created when installing
+ESXi to log into the interface. After logging in, go to *Storage* to
```
wblock: s/into/in to/

userguide/install.rst
```diff
@@ -999,115 +999,91 @@ option. You can permanently boot from disk by removing the
...
 
+   .. _esxi ready to complete:
 
-Click :guilabel:`Next` and :guilabel:`Finish`. The new virtual machine
-is listed in the left frame. Right-click the virtual machine and
-select :guilabel:`Edit Settings` to access the screen shown in
-:numref:`Figure %s <esxi_vm_settings_fig>`.
+   .. figure:: images/esxi_ready_complete.png
 
+Additional disks can be added to a VM. Right-click the VM and click
```
wblock: "Additional" and "added" are redundant.  This can just be a "To add more disks to a VM, ..."

PR #778, opened 2019-03-14
71619: Remove zvol snapshot delete entry from intro.rst.

Mrt134:
From Dru's feedback.
HTML build test: no issues

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1149,8 +1149,8 @@ to access the :guilabel:`Delete zvol`, :guilabel:`Edit Zvol`,
 :guilabel:`Create Snapshot`, and, for an existing zvol snapshot,
 :guilabel:`Promote Dataset` options.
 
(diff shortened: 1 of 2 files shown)


userguide/storage.rst
```diff
@@ -1149,8 +1149,8 @@ to access the :guilabel:`Delete zvol`, :guilabel:`Edit Zvol`,
 :guilabel:`Create Snapshot`, and, for an existing zvol snapshot,
 :guilabel:`Promote Dataset` options.
 
-Similar to datasets, a zvol name cannot be changed. Deleting a zvol
-also removes all snapshots of that zvol.
+Similar to datasets, a zvol name cannot be changed. Choosing to delete a
+zvol warns that all snapshots of that zvol will also be deleted.
```
wblock: "Choosing to delete" is unclear here.  Also, this seems like it should be a separate paragraph.
"Choosing a zvol for deletion shows a warning that all snapshots of that zvol will also be deleted."

PR #780, opened 2019-03-14
80679 & 77089: Mention init/shutdown scripts use 'sh' not the default 'zsh'

a-stjohn:


userguide/tasks.rst
```diff
@@ -184,6 +184,9 @@ path to the command is shown:
 When scheduling a script, test the script first to verify it is
 executable and achieves the desired results.
 
+.. note:: The default %brand% shell is :file:`zsh`, but
+   init/shutdown scripts are run using :file:`sh`.
```
wblock: Not sure if mentioning the default shell is necessary or even wise to do here.
```suggestion
```

userguide/tasks.rst
```diff
@@ -184,6 +184,9 @@ path to the command is shown:
 When scheduling a script, test the script first to verify it is
 executable and achieves the desired results.
 
+.. note:: The default %brand% shell is :file:`zsh`, but
```
wblock: I don't think we should even mention the default interactive shell here, just which one is used for scripts.  sh is a command, so uses different markup.  Please build-test this.

```suggestion
.. note:: Init/shutdown scripts are run with :command:`sh`.
```

dlavigne:
Master equivalent: https://github.com/freenas/freenas-docs/commit/d6847b695895a152c3c3f8a71c9b42bc2ba4c228

PR #784, opened 2019-03-14
69186: Expand LUN ID description

Mrt134:
From notes in Redmine ticket.
HTML build test: no issues.

### DIFF
--- a/userguide/sharing.rst
+++ b/userguide/sharing.rst
@@ -2176,10 +2176,8 @@ and extents.
    |             |                |                                                           |
    +-------------+----------------+-----------------------------------------------------------+
    | LUN ID      | integer        | Select or enter a value between *0* and *1023*. Some      |
-   |             |                | initiators expect a value less than *256*. Each target    |
-   |             |                | must have a unique :guilabel:`LUN ID`, but associated     |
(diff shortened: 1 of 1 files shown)


userguide/sharing.rst
```diff
@@ -2168,19 +2168,22 @@ and extents.
...
+   +-------------+----------------+-----------------------------------------------------------+
+   | Setting     | Value          | Description                                               |
+   |             |                |                                                           |
+   +=============+================+===========================================================+
+   | Target      | drop-down menu | Select an existing target.                                |
+   |             |                |                                                           |
+   +-------------+----------------+-----------------------------------------------------------+
+   | LUN ID      | integer        | Select or enter a value between *0* and *1023*. Some      |
+   |             |                | initiators expect a value less than *256*. Each target    |
+   |             |                | must have a unique :guilabel:`LUN ID`, but associated     |
```
wblock: "Use unique LUN IDs for each associated target."

wmohanlon: "associated targets with unique targets" doesn't sound quite right.
How about something like "LUN IDs must be unique to each target"  ?

wmohanlon: Wait, that's what wblock is saying.

PR #58, opened 2018-03-21
Removed link to deleted blog post.

sambonham:

PR #788, opened 2019-03-15
Doc Queue #58: Add note to vmware instructions about configuring jail networking

Mrt134:
HTML build test: no issues.


Mrt134:
Waiting to merge until QE can confirm instruction.

userguide/install.rst
```diff
@@ -1114,3 +1114,10 @@ click :menuselection:`Options --> Advanced --> File Locations`.
 Locate the path for the Configuration file named :file:`filename.vmx`.
 Open that file in a text editor, change :guilabel:`hpet0.present` from
 *true* to *false*, and save the change.
+
+If plugins or jails inside the %brand% VM are encountering network
```
a-stjohn: This is a lengthy "if-then" sentence. It may be fine in this case though.

wblock: It's not terrible with just the link text, the URLs make it look much longer.

userguide/install.rst
```diff
@@ -1114,3 +1114,10 @@ click :menuselection:`Options --> Advanced --> File Locations`.
...
+connectivity issues, edit the connected
```
wblock: Any connectivity issues, or just specific ones?

PR #789, opened 2019-03-15
Doc Queue #58: Add note to vmware instructions about configuring jail networking

Mrt134:
HTML build test: no issues.


Mrt134:
Waiting to merge until QE can confirm instruction.

userguide/install.rst
```diff
@@ -1118,3 +1118,10 @@ click :menuselection:`Options --> Advanced --> File Locations`.
 Locate the path for the Configuration file named :file:`filename.vmx`.
 Open that file in a text editor, change :guilabel:`hpet0.present` from
 *true* to *false*, and save the change.
+
+If plugins or jails inside the %brand% VM are encountering network
+connectivity issues, edit the connected
```
wblock: It would be better to tell the user what they are doing first, then tell them where. So:
"connectivity issues, allow MAC spoofing and enable promiscuous mode in the"

wblock: I still think that "connectivity issues" is too vague and should be better defined, but understood if we don't have that information.

PR #804, opened 2019-04-08
NAS-101187: Update guidance about moving the system dataset to an enc…

Mrt134:
…rypted volume.

Rework the note box into the surrounding text and describe the conditions for using an encrypted volume for the system dataset.
HTML build test: no issues.

Not ready for porting to master or angular branches.


userguide/system.rst
```diff
@@ -974,19 +974,25 @@ controller users and groups.
...
 system log and :ref:`Reporting` information. If there are lots of log
 entries or reporting information, moving these to the system dataset
 will prevent :file:`/var/` on the device holding the operating system
 from filling up as :file:`/var/` has limited space.
 
-Use the drop-down menu to select the ZFS volume (pool) to contain the
-system dataset. Whenever the location of the system dataset is
-changed, a pop-up warning indicates that the SMB service must be
-restarted, causing a temporary outage of any active SMB connections.
+Use the :guilabel:`System dataset pool` drop-down menu to select the ZFS
```
```suggestion
Use the :guilabel:`System dataset pool` drop-down menu to select the
```

wblock: The system dataset can be moved to unencrypted volumes (pools) or encrypted volumes which do not have passphrases. If the system dataset is moved to an encrypted volume, that volume is no longer allowed to be locked or have a passphrase set.

userguide/system.rst
```diff
@@ -974,19 +974,25 @@ controller users and groups.
...
+volume (pool) to contain the system dataset.
+
+The system dataset can be moved to an encrypted volume, provided that
+volume has no configured passphrase. Encrypted, locked volumes are not
+displayed in the :guilabel:`System dataset pool` drop-down menu. An
+encrypted volume containing the system dataset cannot be locked or have
+a passphrase created for that volume.
+
+Whenever the location of the system dataset is changed, a pop-up warning
```
wblock: This sentence makes it sound like the pop-up causes the outage, but it doesn't.
Should use "dialog" instead of "pop-up", too.

PR #814, opened 2019-04-09
NAS-100097: Update "Show Text Console without Password Prompt" descri…

Mrt134:
…ption.

HTML build test: no issues.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -579,10 +579,8 @@ The configurable settings are summarized in
    | Setting                    | Value               | Description                                                                   |
    |                            |                     |                                                                               |
    +============================+=====================+===============================================================================+
-   | Show Text Console          | checkbox            | Set for the system to immediately display the |web-ui| address and text       |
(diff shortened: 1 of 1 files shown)


userguide/system.rst
```diff
@@ -579,8 +579,9 @@ The configurable settings are summarized in
    | Setting                    | Value               | Description                                                                   |
    |                            |                     |                                                                               |
    +============================+=====================+===============================================================================+
-   | Show Text Console          | checkbox            | Set for the system to immediately display the text console after booting.     |
-   | without Password Prompt    |                     | Unset to require logging into the system before the console menu is shown.    |
+   | Show Text Console          | checkbox            | Set for the system to immediately display the |web-ui| address and text       |
```
wblock: It's confusing that |web-ui| address (should be "IP address") is mentioned in both cases.  If it shows that in both cases, it would be less confusing to just not mention it.
"Set for the text console to be available without entering a password."

PR #817, opened 2019-04-09
Nas 101056 Fix text that was refereing to USB devices other than for install only.

a-stjohn:
Cherry picked from PR #816


a-stjohn:
Closing the PR. Massively confused on what information was supposed to make it through the cherry-pick and was not. Stuff is messed up.

PR #827, opened 2019-04-09
NAS-100155 Add section and docs for new task manager in GUI.

a-stjohn:
(cherry picked from commit 4f39c032e065e2128166114f91a95d2c04bef3e2)


dlavigne:
Is this the correct branch for an 11.3-BETA1 ticket?

a-stjohn:
> Is this the correct branch for an 11.3-BETA1 ticket?

@dlavigne I thought this feature was for both 11.2 and 11.3-BETA1. If it's ONLY for 11.3-BETA1, then the correct branch is `angulargui`.

dlavigne:
According to the ticket, it is for BETA1.

wblock:
Not needed, change is only in 11.3-BETA1.

PR #825, opened 2019-04-09
Work with Support to refine SED documentation.

Mrt134:
cherry-pick PR #824 to 11.2-angular.

PR #828, opened 2019-04-10
Work with Support to refine SED docs

Mrt134:
Cherry-pick previous commits to master and rework instructions to match legacy UI.
HTML build test: no issues.


Mrt134:
Reviewed in PR #825

PR #845, opened 2019-04-11
NAS-100393: Add text explaining system virtualization error.

Mrt134:
HTML build test: no issues


userguide/vms.rst
```diff
@@ -20,6 +20,8 @@ other VMs.
 virtual machine software. This type of virtualization requires an
 Intel processor with Extended Page Tables (EPT) or an AMD processor
 with Rapid Virtualization Indexing (RVI) or Nested Page Tables (NPT).
+%brand% shows an error during VM creation when these features are not
```
wblock: Be more specific about when the error is shown:
"VMs cannot be created unless the host system supports these features."

PR #848, opened 2019-04-11
NAS-100376: Add "Exclude" to Cloud Sync Options table

Mrt134:
Also add entry to intro
HTML build test: no issues.


userguide/tasks.rst
```diff
@@ -168,6 +168,9 @@ shows the configuration options for Cloud Syncs.
    | Enabled                | checkbox            | Unset to temporarily disable this Cloud Sync.                                                          |
    |                        |                     |                                                                                                        |
    +------------------------+---------------------+--------------------------------------------------------------------------------------------------------+
+   | Exclude                | string              | Newline-separated list of files and directories to exclude from sync. See                              |
```
wblock:
```suggestion
   | Exclude                | string              | List of files and directories to exclude from sync, one per line. See```

PR #850, opened 2019-04-11
NAS-100394: Add "Exclude" to Cloud Sync Options table

Mrt134:
Also add entry to intro
Identical to #848 
HTML build test: no issues.


Mrt134:
Reviewed in #848 .

PR #852, opened 2019-04-11
NAS-100338 Added information about the 'VERIFY CREDENTIAL' button.

a-stjohn:

PR #853, opened 2019-04-11
NAS-100461 Added information about the 'VERIFY CREDENTIAL' button.

a-stjohn:


dlavigne:
This one has a grammo and does not match https://github.com/freenas/freenas-docs/pull/852/files. We'll also need an intro entry for both angulargui and master since this is a new button.

a-stjohn:
> This one has a grammo and does not match https://github.com/freenas/freenas-docs/pull/852/files. We'll also need an intro entry for both angulargui and master since this is a new button.

Fixed.

PR #858, opened 2019-04-11
NAS-100496  Updated guide to reflect the removal of "Protocol" field in System > General

a-stjohn:


a-stjohn:
This is marked `don't merge yet` because the new and the old UI differ. In the old UI the `Protocol` field is gone and no long in use. In the new UI the protocol field is present, but it doesn't seem to work or matter that it is there. I've spoke with Erin. She believes that there should be a ticket about this.

PR #872, opened 2019-04-15
SED section: Work mav's comments into SED specs list.

Mrt134:

(cherry-pick from #869 )


userguide/system.rst
```diff
@@ -729,15 +729,34 @@ Self-Encrypting Drives
...
 
-* TCG Enterprise standard for newer enterprise-grade SAS devices
+* `TCG Opalite <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Opalite_SSC_FAQ.pdf>`__
+  is a reduced form of OPAL 2
+
+* `TCG Pyrite Version 1 <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Pyrite_SSC_v1.00_r1.00.pdf>`__
+  and
+  `TCG Pyrite Version 2 <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Pyrite_SSC_v2.00_r1.00_PUB.pdf>`__
+  provide a logical equivalent of the legacy ATA security for non-ATA
+  devices. Hardware encryption is removed and only the drive firmware is
```
amotin: I'd write here something like: "It is a version of Opalite with hardware encryption is removed ...".

userguide/system.rst
```diff
@@ -729,15 +729,37 @@ Self-Encrypting Drives
...
+  `TCG Pyrite 2 <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Pyrite_SSC_v2.00_r1.00_PUB.pdf>`__
+  provide a logical equivalent of the legacy ATA security for non-ATA
+  devices. Hardware encryption is removed and only the drive firmware is
+  used to protect the device.
+
+  .. warning:: Pyrite Version 1 does not have PSID support and **can
+     become unusable if the password is lost.**
+
+* `TCG Enterprise <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-SSC_Enterprise-v1.01_r1.00.pdf>`__
+  for high performance systems
```
amotin: I am not sure how this is related to performance.  I'd say more to systems with many data disks, not used for booting, since it does not provide mechanisms to ask for for unlocking password before OS boot.

Mrt134:
Ok, updated again with your comments. Thanks for your review!

userguide/system.rst
```diff
@@ -729,15 +729,41 @@ Self-Encrypting Drives
...
+  devices. Only the drive firmware is used to protect the device.
+
+  .. danger:: Pyrite Version 1 SEDs do not have PSID support and **can
+     become unusable if the password is lost.**
+
+
+* `TCG Enterprise <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-SSC_Enterprise-v1.01_r1.00.pdf>`__
+  is designed for systems with many data disks. **Do not use as a boot
+  device!** TCG Enterprise SEDs cannot be unlocked before the operating
+  system boots.
```
amotin: I don't think it should be so aggressively mentioned. It is just a fact of what those disks are, and we in no way advertise or use SED on boot disks now.

Mrt134: Ok, understood. I'll adjust as needed.

PR #888, opened 2019-04-16
NAS-100631: Add text about certificate delete change

Mrt134:
HTMl build test: no issues.


Mrt134:
Closed after discussing with Warren.

PR #919, opened 2019-04-18
NAS-100947 Add docs for "Enable Shadow Copies" checkbox.

a-stjohn:


userguide/sharing.rst
```diff
@@ -952,6 +949,11 @@ provides more details for each configurable option.
    |                                |               |          | space and is recreated whenever a file is deleted.                                                                                                   |
    |                                |               |          |                                                                                                                                                      |
    +--------------------------------+---------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------+
+   | Enable Shadow Copies           | checkbox      | ✓        | Export ZFS snapshots as                                                                                                                              |
```
wblock:
``` Expose ZFS snapshots as
 `Windows Shadow Copies <>`__. ```
VSS stands for both Visual Source Safe and Volume Shadow Copies Service.  Here, it is the second one, and the link should go to that.  But I didn't find a link for it in a quick search.

PR #925, opened 2019-04-18
Update network reset message

wblock:
NAS-101226


dlavigne:
This one will also need the TN section as the new UI will be an option in 11.3.

PR #952, opened 2019-04-22
NAS-100811: Rework jail update/upgrade and plugin update/upgrade sections:

Mrt134:
Rename plugin update subsection and replace upgrade instructions with a ref to jail upgrade section.
Screenshot updating a plugin and uncomment the image in the text.
Rework the text in jails/update and upgrade section.


userguide/intro.rst
```diff
@@ -155,6 +155,9 @@ These screen options have changed:
   :menuselection:`Network --> Interfaces --> ADD` has been renamed
   to :guilabel:`Description`.
 
+* The :guilabel:`Upgrade` option has been removed from
+  :menuselection:`Plugins --> Installed -->\ ` |ui-options|.
```
wblock: See if escape is necessary here.

Mrt134: nope, looks like it can be removed

PR #982, opened 2019-04-24
NAS-101373 Update all the redmine links to point to Jira.

a-stjohn:
fixed


wblock:
This is an ideal opportunity for a replacement.

PR #1044, opened 2019-05-01
NAS-100572: Update Network/Global Config/Hostname (Virtual) descripti…

Mrt134:
…on to include allowed characters.
TrueNAS HTML build test: no issues.
cherry-pick from #1043


userguide/network.rst
```diff
@@ -94,7 +94,8 @@ but can be changed to meet requirements of the local network.
    |                         |               |                                                                                                   |
    +-------------------------+---------------+---------------------------------------------------------------------------------------------------+
    | Hostname (Virtual)      | string        | Virtual host name. When using a virtualhost, this is also used as the Kerberos principal name.    |
```
a-stjohn: Reorder so the comma isn't there.

```suggestion
   | Hostname (Virtual)      | string        | Virtual host name. This is also used as the Kerberos principal name when using a virtualhost.    |
```

PR #1045, opened 2019-05-01
Add security issue info when using "allow_raw_sockets".

a-stjohn:


userguide/jails.rst
```diff
@@ -488,6 +489,16 @@ of these options.
    +--------------------------+--------------+---------------------------------------------------------------------------------------------------------------------+
 
 
+.. warning::
+   :name: security warning
+
+   :guilabel:`allow_raw_sockets` enables Inernet Control Message
```
wblock: Typo.

userguide/jails.rst
```diff
@@ -488,6 +489,16 @@ of these options.
...
+   Protocol (ICMP) and
```
wblock: It doesn't strictly allow tcpdump, but the use of tcpdump or any other tools that require raw sockets, like nmap.

userguide/jails.rst
```diff
@@ -488,6 +489,16 @@ of these options.
...
+   `tcpdump <https://en.wikipedia.org/wiki/Tcpdump>`__
+   in the jail. This allows access to the raw sockets of the host and
+   poses a security threat.
```
wblock: s/poses/can pose/, but "pose" is a strange verb to use.  How about "can be a security hazard."?
We should also link to something that describes raw sockets more, like https://en.wikipedia.org/wiki/Network_socket#Datagram_socket.

userguide/jails.rst
```diff
@@ -430,7 +430,8 @@ of these options.
    |                          |              | `ping(8) <https://www.freebsd.org/cgi/man.cgi?query=ping>`__ and                                                    |
    |                          |              | `traceroute(8) <https://www.freebsd.org/cgi/man.cgi?query=traceroute>`__                                            |
    |                          |              | require raw sockets to operate inside a jail. When set, the source IP addresses are enforced to comply with the     |
```
wblock: Not at all clear what "the source IP addresses are enforced to comply with the IP address bound to the jail" means.  "Enforced to comply" in particular.

PR #1061, opened 2019-05-09
Revert "NAS-100716: Remove TrueNAS note about alert if system dataset is not …"

Mrt134:
Reverts freenas/freenas-docs#1046


dlavigne:
Why was this reverted? It was correct as-is and now fails testing.

PR #1093, opened 2019-05-21
NAS-100410 : Add NVDIMM alert to alertevents.rst

Mrt134:
TrueNAS only
TrueNAS HTML build test: no issues.

### DIFF
--- a/userguide/snippets/alertevents.rst
+++ b/userguide/snippets/alertevents.rst
@@ -112,7 +112,7 @@ Some of the conditions that trigger an alert include:
 * the interface which is set as critical for failover is not found
   or is not configured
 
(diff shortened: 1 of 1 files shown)


userguide/snippets/alertevents.rst
```diff
@@ -112,6 +112,8 @@ Some of the conditions that trigger an alert include:
 * the interface which is set as critical for failover is not found
   or is not configured
 
+* NVDIMM problems are detected
```
a-stjohn: The introductory sentence before this list is "Some of the conditions that trigger an alert include:" Thus, 
```suggestion
* NVDIMM problems.
```
would make more sense.

PR #1108, opened 2019-05-23
 NAS-101234: Add text about CARP status alert to system.rst and alertevents.rst

Mrt134:
TrueNAS HTML build: no issues.

Warning: I couldn't find the relevant alert in our TrueNAS 11.2 testing system. I'm trying to track it down but haven't had any luck yet. The correct alert is visible in a TN 11.3 system.


userguide/system.rst
```diff
@@ -2834,6 +2834,10 @@ in standby mode. The |web-ui| shows an additional
 :guilabel:`Force Takeover` button which can be used to force that node
 to take control.
 
+Failover is non-functional when both nodes have an identical CARP
```
wblock: Didn't we fix this wording earlier in another PR?  Being specific about whether failover just will not work or is not allowed, clarifying that it's not that they have an identical CARP state, but the *same* CARP state...

PR #1109, opened 2019-05-23
 NAS-101234: Add text about CARP status alert to system.rst and alertevents.rst

Mrt134:
TrueNAS HTML build test: no issues
cherry-pick from #1108


Mrt134:
Branch updated. Also used the proposed replacement |ctrlrs-term| from #1136 .

PR #1110, opened 2019-05-23
redmine 27384 fix (master)

a-stjohn:


userguide/storage.rst
```diff
@@ -1767,11 +1767,11 @@ in :ref:`Encryption` **before** attempting to replace the failed
...
 3, a prompt will appear to input and confirm the passphrase for the
 pool. Enter this information then click :guilabel:`Replace Disk`.
-Immediately
-:ref:`restore the encryption keys to the pool<Managing Encrypted Volumes>`.
 
-.. warning:: Access to the pool will be permanently lost unless the
-   encryption keys are restored to the pool before the next system reboot!
+Wait until the resilvering is complete before
+:ref:`restoring the encryption keys to the pool<Managing Encrypted Volumes>`.
+**If the encryption keys are not restored before the next reboot,
```
wblock: This changes the sense of the sentence. The original was not a "might", but a "will".  The new version implies that they might get away with not restoring the keys, while the original says it will definitely go wrong if the keys are not restored before a reboot.

userguide/storage.rst
```diff
@@ -1767,11 +1767,11 @@ in :ref:`Encryption` **before** attempting to replace the failed
 drive. Then, follow the steps 1 and 2 as described above. During step
 3, a prompt will appear to input and confirm the passphrase for the
 pool. Enter this information then click :guilabel:`Replace Disk`.
-Immediately
-:ref:`restore the encryption keys to the pool<Managing Encrypted Volumes>`.
 
-.. warning:: Access to the pool will be permanently lost unless the
-   encryption keys are restored to the pool before the next system reboot!
+Wait until the resilvering is complete before
+:ref:`restoring the encryption keys to the pool<Managing Encrypted Volumes>`.
```
wblock: Have you built this?  I'm not sure if the space between the link text and the URL is required or not.

PR #1135, opened 2019-05-29
NAS-100122: Replace all textual uses of the terms "node" and "storage controller"

Mrt134:
Register new replacements in conf.py
Search through the guide and replace every relevant instance of "node", "storage controller", and variants of the terms.
Current replacement term setting: "storage controller" and variants
TrueNAS html build testing: no related warnings
FreeNAS html build testing: no issues

### DIFF
--- a/userguide/conf.py
+++ b/userguide/conf.py
@@ -78,15 +78,16 @@ rst_prolog = u'''
 .. |dockerhost|           replace:: Docker VM
 .. |help-pin|             replace::  (Pin)
 .. |help-unpin|           replace::  (Unpin)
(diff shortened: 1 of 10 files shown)


userguide/conf.py
```diff
@@ -78,6 +78,15 @@
 .. |dockerhost|           replace:: Docker VM
 .. |help-pin|             replace::  (Pin)
 .. |help-unpin|           replace::  (Unpin)
+.. |Node|                 replace:: Storage Controller
```
wblock: As per our meeting, use `ctrlr-term`, `Ctrlr-term`, `ctrlrs-term`, and so on.

PR #1137, opened 2019-05-30
NAS-101494: Add noacl vfs module to sharing.rst "Available VFS Module…

Mrt134:
…s" table

- HTML build test: no issues.


dlavigne:
This PR is not needed as option was added for 11.2.

PR #1174, opened 2019-06-10
NAS-100021: Add one-time replication example

Mrt134:
This is DRAFT text of the current procedure. This is still in testing and requires further review and refinement to stay in sync with replication system changes.


userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
 replication task.
 
 
+.. _One-time Replication:
+
+One-time Replication
```
```suggestion
One-Time Replication
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
 replication task.
 
 
+.. _One-time Replication:
```
```suggestion
.. _One-Time Replication:
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+
+One-time replications copy manually created snapshots of a dataset to
+another target dataset. This kind of replication is compatible with any
+replication :guilabel:`Transport` method. To configure a one-time
+replication, go to
+:menuselection:`Storage --> Replication Tasks`
+and click :guilabel:`Add Replication`.
+
+Enter a descriptive :guilabel:`Name`, choose the :guilabel:`Direction`,
+and configure any :guilabel:`Transport` settings.
```
wblock: "any" is confusing here.  "any desired" is clearer.

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+
+Enter the :guilabel:`Source Datasets` that have snapshots to be copied
```
```suggestion
Enter the :guilabel:`Source Datasets` with snapshots to be copied
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+to the target.
+
+Enter the :guilabel:`Target Dataset` that will store the replicated
```
```suggestion
Enter the :guilabel:`Target Dataset` where replicated snapshots will be stored
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+snapshots and decide if child dataset snapshots should also be
```
```suggestion
and decide whether child dataset snapshots should also be
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+replicated.
+
+Do not select a periodic snapshot task for one-time replications. Enter
+a :guilabel:`Also include naming schema` to rename replicated snapshots
```
wblock: "Enter a"?  Is the "a" extra here?

This is also confusing.  Tell what the naming schema is (bad name, maybe we should just call it a naming format?) and then say it can be used to rename snapshots, with this example doing it based on replication time. (replication time, or snapshot time?)
"The naming schema can be used to rename replicated snapshots

userguide/storage.rst
```diff
@@ -2431,6 +2427,43 @@ The remaining settings are left at the default choices. Make sure
...
+according to the replication time. This requires including the *%Y*,
+*%m*, *%d*, *%H*, and *%M* strings. This prevents replication errors
+from duplicate snapshot names from occurring.
```
wblock: Not clear.  "Renaming replicated snapshots helps to avoid having multiple snapshots with the same name, (causing what? replication to fail?)"

userguide/storage.rst
```diff
@@ -2431,6 +2427,44 @@ The remaining settings are left at the default choices. Make sure
 replication task.
 
 
+.. _One-Time Replication:
+
+One-Time Replication
+^^^^^^^^^^^^^^^^^^^^
+
+One-time replications copy manually created snapshots of a dataset to
```
```suggestion
One-time replications copy manually-created snapshots of a dataset to
```

userguide/storage.rst
```diff
@@ -2431,6 +2427,44 @@ The remaining settings are left at the default choices. Make sure
...
+target.
+
+Enter the :guilabel:`Target Dataset` where replicated snapshots will be
+stored and decide whether child dataset snapshots should also be
+replicated.
+
+Do not select a periodic snapshot task for one-time replications. To
+replicate a single snapshot, enter the name of the snapshot in
+:guilabel:`Also include naming schema`. Replace the year, month, day,
+hour, and minute values with *%Y*, *%m*, *%d*, *%H*, and *%M* strings.
```
wblock: These should use :literal: tags, since they are literal values.

This seems really confusing for the user, though.  They have to enter a pattern match to duplicate a single snapshot?

Mrt134: That does appear to be the current behavior.

PR #1194, opened 2019-06-17
NAS-101051 comment out all of docker docs

a-stjohn:
NOTE: I decided to comment out the docker section instead of deleting it. This is because docs still need to be written on how to get a docker vm running (since the docker vm type was removed from the UI) and I may be able to use information from the section.


a-stjohn:
UPDATE: I no longer need to write docs on how to get a docker instance running in a VM. I have removed the docker docs in PR #1254 .

PR #1202, opened 2019-06-21
NAS-102188: Rework dataset delete instructions and add intro entry

Mrt134:
HTML test: no issues.

PR #1203, opened 2019-06-21
NAS-102188: Update dataset delete description and add warning

Mrt134:
HTML test: no issues.
cherry-pick #1202 
Rework changes to match legacy UI.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1410,9 +1410,15 @@ the origin filesystem becomes a clone of the clone making it possible
 to destroy the filesystem that the clone was created from. Otherwise,
 a clone cannot be deleted while the origin filesystem exists.
 
-**Destroy Dataset:** clicking the :guilabel:`Destroy Dataset` button
-causes the browser window to turn red to indicate that this is a
-destructive action. Clicking :guilabel:`Yes` proceeds with the deletion.
+**Destroy Dataset:** removes the dataset, snapshots of that dataset, and
+any objects stored within the dataset. Clicking
+:guilabel:`Destroy Dataset` causes the browser window to turn red to
+indicate that this is a destructive action. To delete the datset, click
+:guilabel:`Yes`.
+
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -1410,9 +1410,15 @@ the origin filesystem becomes a clone of the clone making it possible
...
-**Destroy Dataset:** clicking the :guilabel:`Destroy Dataset` button
-causes the browser window to turn red to indicate that this is a
-destructive action. Clicking :guilabel:`Yes` proceeds with the deletion.
+**Destroy Dataset:** removes the dataset, snapshots of that dataset, and
+any objects stored within the dataset. Clicking
+:guilabel:`Destroy Dataset` causes the browser window to turn red to
+indicate that this is a destructive action. To delete the datset, click
+:guilabel:`Yes`.
+
+.. danger:: Destroying a dataset is a permanent action and can result in
```
a-stjohn: Might be a question for Warren, but do we want to be extra sure here? That is, indicate that it surely causes data loss...?

```suggestion
.. danger:: Destroying a dataset is a permanent action and results in
```

userguide/storage.rst
```diff
@@ -1410,9 +1410,15 @@ the origin filesystem becomes a clone of the clone making it possible
...
-**Destroy Dataset:** clicking the :guilabel:`Destroy Dataset` button
-causes the browser window to turn red to indicate that this is a
-destructive action. Clicking :guilabel:`Yes` proceeds with the deletion.
+**Destroy Dataset:** removes the dataset, snapshots of that dataset, and
+any objects stored within the dataset. Clicking
+:guilabel:`Destroy Dataset` causes the browser window to turn red to
+indicate that this is a destructive action. To delete the datset, click
+:guilabel:`Yes`.
+
+.. danger:: Destroying a dataset is a permanent action and results in
```
wblock: Warn first, then give instructions on how to do it.

PR #1227, opened 2019-06-26
NAS-101108: Review ported SSH Connections text and update for legacy ui

Mrt134:
HTML build test: no issues.
Run through examples and verify they are still valid.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -1526,7 +1526,7 @@ and click :guilabel:`Add SSH Connection`.
    |                 |                | |web-ui|, go to :menuselection:`Storage --> Replication Tasks`, and click           |
    |                 |                | :guilabel:`Temporary Auth Token`.                                                   |
    +-----------------+----------------+-------------------------------------------------------------------------------------+
-   | Username        | string         | User account name to use for logging in to the remote system                        |
+   | Username        | string         | User account name to use for logging in to the remote system.                       |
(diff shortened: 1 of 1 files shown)


userguide/system.rst
```diff
@@ -1519,18 +1518,21 @@ and click |ui-add|.
...
    | FreeNAS URL     | string         | Hostname or IP address of the remote %brand% system. Only available                 |
    |                 |                | with *Semi-automatic* configurations. A valid URL scheme is required. Example:      |
-   |                 |                | :samp:`https://{10.231.3.76}`                                                       |
+   |                 |                | :samp:`http://{10.231.3.76}`                                                        |
    +-----------------+----------------+-------------------------------------------------------------------------------------+
-   | Username        | string         | User account name to use for logging in to the remote system                        |
+   | Auth Token      | string         | Added when *Semi-automatic* is the :guilabel:`Setup Method`. Temporary access token |
+   |                 |                | created by the remote %brand% system. To obtain this token, log in to the           |
+   |                 |                | %brand% legacy                                                                      |
```
a-stjohn: Do we need "legacy" here? This is the master branch anyway...

Mrt134: I wrote it like this because the default in 11.3 is to show logging in to the new UI, so thought it could help prevent confusion to make the distinction. Also, that specific button in Storage/Replication Tasks is only visible in the legacy UI. If you think it's unnecessary, I can adjust it.

PR #1229, opened 2019-06-26
NAS-101110: Review UI fields that are part of Replication tasks

Mrt134:
HTML build test: no issues.


Mrt134:
Includes removing text about the "Write_EMBEDDED" field in NAS-102044.

PR #1237, opened 2019-06-27
NAS-101538 add docs for not using space in task schedulers.

a-stjohn:


userguide/tasks.rst
```diff
@@ -37,6 +37,8 @@ Each of these tasks is described in more detail in this section.
    automatically-created task. :ref:`S.M.A.R.T. Tests` and
    :ref:`Periodic Snapshot Tasks` must be set up manually.
 
+.. note:: Spaces cannot be used in the task scheduler.
```
Mrt134: Probably would be best to avoid having two note boxes right next to each other.

a-stjohn: Ya that crossed my mind. I couldn't really think of another solution for such a small aside about the scheduler. Do you have suggestions?

wblock: This is probably not specific enough.  Spaces cannot be used at all, or only in certain fields, or not in times?  This note might be better with the specific fields where spaces are not allowed, or integrated into the descriptions of each field that does not allow spaces.

userguide/tasks.rst
```diff
@@ -88,7 +89,7 @@ lists the configurable options for a cron job.
    |                     |                             |                                                                                                         |
    +---------------------+-----------------------------+---------------------------------------------------------------------------------------------------------+
    | Schedule a Cron Job | drop-down menu              | Select how often to run the cron job. Choices are *Hourly*, *Daily*, *Weekly*, *Monthly*, or *Custom*.  |
-   |                     |                             | Select *Custom* to open the advanced scheduler.                                                         |
+   |                     |                             | Select *Custom* to open the advanced scheduler. Spaces are not allowed in the custom scheduler.         |
```
wblock: This is better, but still not specific.  Are spaces not allowed in the name, in the time values, where?

userguide/tasks.rst
```diff
@@ -1562,6 +1568,7 @@ screen.
    +----------------+-----------------------------+-------------------------------------------------------------------------------------------------------------+
    | Schedule the   | drop-down menu              | Choose how often to run the scrub task. Choices are *Hourly*, *Daily*, *Weekly*, *Monthly*, or *Custom*.    |
    | Scrub Task     |                             | Select *Custom* to open a visual scheduler for selecting minutes, hours, days, month, and days of week.     |
+   |                |                             | Spaces are not allowed in the custom scheduler.                                                             |
```
wblock: Missed one.

PR #1240, opened 2019-06-28
NAS-100703 docs for cloud sync fail alert

a-stjohn:


userguide/snippets/alertevents.rst
```diff
@@ -82,6 +82,8 @@ Some of the conditions that trigger an alert include:
 
 * a periodic snapshot or replication task fails
 
+* a cloud sync task fails
```
Mrt134: Would a :ref: to cloud sync tasks be useful here?

PR #1243, opened 2019-07-01
NAS-102277 remove docs for theme selector tool.

a-stjohn:


dlavigne:
Shouldn't this be a move rather than a remove?

a-stjohn:
> Shouldn't this be a move rather than a remove?

It didn't appear so. The theme selector wasn't present in preferences. However, the preferences already had an option to choose between themes. It looks like it would've been redundant it it was moved there.

PR #1253, opened 2019-07-02
NAS-100911: Add b2-chunk-size field docs

Mrt134:
Port existing help text and rework slightly for the guide.
HTML build test: no issues.


userguide/tasks.rst
```diff
@@ -1670,6 +1670,10 @@ shows the configuration options for Cloud Syncs.
    |                     |                     | :guilabel:`Container`: Only appears when a :literal:`AZUREBLOB` credential is selected for the          |
    |                     |                     | :guilabel:`Credential`. Enter the name of the pre-configured Microsoft Azure Blob container.            |
    +---------------------+---------------------+---------------------------------------------------------------------------------------------------------+
+   | Upload Chunk Size   | integer             | Only appears with a *(B2)* :guilabel:`Credential`. Files are split into chunks of this size before      |
```
a-stjohn: As Warren would say: What it does first, then other information (like "Only appears with...") last.

PR #1261, opened 2019-07-02
NAS-100959: new alert when cloud sync task fails.

Mrt134:
HTML build test: no issues.


Mrt134:
Two different tickets in Jira for the same thing?

PR #1270, opened 2019-07-03
NAS-102231: Fix up hubiC text

Mrt134:


userguide/system.rst
```diff
@@ -1222,10 +1222,13 @@ Cloud Credentials
 %brand% can use cloud services for features like :ref:`Cloud Sync`.
 The credentials to provide secure connections with cloud services
 are entered here. Amazon S3, Backblaze B2, Box, Dropbox, FTP, Google
-Cloud Storage, Google Drive, HTTP, Hubic, Mega, Microsoft Azure Blob
+Cloud Storage, Google Drive, HTTP, hubiC, Mega, Microsoft Azure Blob
 Storage, Microsoft OneDrive, pCloud, SFTP, WebDAV, and Yandex are
 supported.
 
+.. note:: The hubiC cloud service has
```
wblock: This should be in the hubiC provider entry, since it only applies to that provider.

PR #1299, opened 2019-07-10
Fix periodic snapshot task and replication task documentation inaccur…

themylogin:
…acies

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1948,9 +1948,9 @@ summarizes the fields in this screen.
    |                   |                   |                                                                                            |
    +-------------------+-------------------+--------------------------------------------------------------------------------------------+
    | Exclude           | string            | Exclude specific child datasets from the snapshot. Use with :guilabel:`Recursive`          |
-   |                   |                   | snapshots. List paths to any child datasets to exclude. Separate multiple entries with a   |
-   |                   |                   | comma (:literal:`,`). Example: :samp:`pool1/dataset1/child1`. A recursive snapshot of      |
-   |                   |                   | :file:`pool1/dataset1` includes all child datasets except :file:`child1`.                  |
+   |                   |                   | snapshots. List names of child datasets to exclude, one per line.                          |
+   |                   |                   | Example: :samp:`pool1/dataset1/child1`. A recursive snapshot of :file:`pool1/dataset1`     |
+   |                   |                   | includes all child datasets except :file:`child1` and it's descendants.                    |
    +-------------------+-------------------+--------------------------------------------------------------------------------------------+
    | Snapshot          | integer and       | Define a length of time to retain the snapshot on this system.                             |
    | Lifetime          | drop-down menu    | After the time expires, the snapshot is removed.                                           |
@@ -1987,8 +1987,7 @@ summarizes the fields in this screen.
 
 If the :guilabel:`Recursive` option is enabled, child datasets of this
 dataset are included in the snapshot and there is no need to create
-snapshots for each child dataset. The downside is that there is no way
-to exclude particular child  datasets from a recursive snapshot.
+snapshots for each child dataset.
 
 Click :guilabel:`SAVE` when finished customizing the task. Defined tasks
 are listed alphabetically in
@@ -2014,10 +2013,10 @@ Replication schedules are typically paired with
 :ref:`Periodic Snapshot Tasks` to generate local copies of important
 data and replicate these copies to a remote system.
 
-Replications require a source system with datset snapshots and a
+Replications require a source system with dataset snapshots and a
 destination that can store the copied data. Remote replications require
-a saved :ref:`SSH Connection <SSH Connections>` on the source system and
-the destination system must be configured to allow :ref:`SSH`
+a saved :ref:`SSH Connection <SSH Connections>` on the local system and
+the remote system must be configured to allow :ref:`SSH`
 connections. Replications to the local system do not require configuring
 SSH.
 
@@ -2063,7 +2062,7 @@ Creating remote, local, %brand% legacy engine, and one-time replications
 are all possible with this screen.
 
 The :guilabel:`Transport` value changes many of the options for
-replication. :ref:`Table %s <zfs_add_replication_task_opts_tab>` uses
+replication. :numref:`Table %s <zfs_add_replication_task_opts_tab>` uses
 shortened versions of the :guilabel:`Transport` methods in the
 :literal:`Transport` column to show which fields appear with the
 different :guilabel:`Transport` options:
@@ -2091,17 +2090,17 @@ different :guilabel:`Transport` options:
    +===========================+===========+================+=================================================================================================================+
    | Name                      | ALL       | string         | Enter a descriptive :guilabel:`Name` for the replication.                                                       |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | Direction                 | ALL       | drop-down menu | Direction of travel. *PUSH* sends snapshots to a destination system. *PULL* receives snapshots from a           |
-   |                           |           |                | destination system. Choosing *PULL* hides the :guilabel:`Periodic Snapshot Tasks` field and renames             |
+   | Direction                 | ALL       | drop-down menu | Direction of travel. *PUSH* sends snapshots to a remote system. *PULL* receives snapshots from a                |
+   |                           |           |                | remote system. Choosing *PULL* hides the :guilabel:`Periodic Snapshot Tasks` field and renames                  |
    |                           |           |                | :guilabel:`Also Include Naming Schema` to :guilabel:`Naming Schema`.                                            |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
    | Transport                 | ALL       | drop-down menu | Method of snapshot transfer:                                                                                    |
    |                           |           |                |                                                                                                                 |
    |                           |           |                | * *SSH* is supported by most systems. A previously-created :ref:`SSH connection <SSH Connections>` is required. |
    |                           |           |                | * *SSH+NETCAT* uses SSH to establish a connection to the destination system, then uses                          |
-   |                           |           |                |   `nc(1) <https://www.freebsd.org/cgi/man.cgi?query=nc>`__ to send an unencrypted data stream for higher        |
+   |                           |           |                |   `py-libzfs <https://github.com/freenas/py-libzfs>`__ to send an unencrypted data stream for higher            |
    |                           |           |                |   transfer speeds. This only works when replicating to a FreeNAS, TrueNAS, or other system with                 |
-   |                           |           |                |   `py-libzfs <https://github.com/freenas/py-libzfs>`__ installed.                                               |
+   |                           |           |                |   py-libzfs installed.                                                                                          |
    |                           |           |                | * *LOCAL* replicates snapshots to another dataset on the same system.                                           |
    |                           |           |                | * *LEGACY* uses the legacy replication engine from %brand% 11.2 and earlier.                                    |
    |                           |           |                |                                                                                                                 |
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -2215,7 +2212,7 @@ different :guilabel:`Transport` options:
    | Number of retries for     | SSH, NCT, | integer        | Number of times the replication is attempted before stopping and marking the task as failed.                    |
    | failed replications       | LOC       |                |                                                                                                                 |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | Logging Level             | ALL       | drop-down menu | Register system messages related to this replication as a specific level in the system log.                     |
+   | Logging Level             | ALL       | drop-down menu | Verbosity of the messages in replication task log.                                                              |
```
```suggestion
   | Logging Level             | ALL       | drop-down menu | Verbosity level of messages in the replication task log.                                                          |
```

userguide/storage.rst
```diff
@@ -2239,8 +2236,8 @@ To see more options for a saved replication, click |ui-options| for that
 task. There are options to :guilabel:`Delete`, :guilabel:`Edit`, or
 immediately start that replication.
 
-When multiple replications have the same schedule, they will run
-serially, one after another. Completion time depends on the number and
+Replications will run in parallel (as long as they don't conflict with each other).
```
wblock: Use present tense, avoid asides, avoid contractions.
```suggestion
Replications run in parallel as long as they do not conflict with each other.
```

userguide/storage.rst
```diff
@@ -2488,7 +2476,8 @@ Replication depends on SSH, disks, network, compression, and
 encryption to work. A failure or misconfiguration of any of these can
 prevent successful replication.
 
-Replication logs are located in :file:`var/log/zettarepl.log`.
+All replication logs are located in :file:`/var/log/zettarepl.log`. You can
```
wblock: Avoid "you" or "your".
```suggestion
Replication logs are saved in :file:`/var/log/zettarepl.log` and can be
```

userguide/storage.rst
```diff
@@ -2488,7 +2476,8 @@ Replication depends on SSH, disks, network, compression, and
...
+inspect logs of any specific replication task by clicking on it's state.
```
wblock: As above about possessives: no apostrophe in possessive "its", but avoid possessives anyway because they are less clear.

Finally, it's not clear what "state" is referring to here.  Probably should use at least `` :guilabel:`State` `` if it is a UI element.

```suggestion
inspected clicking on their state.
```

userguide/storage.rst
wblock: Is it too late to change the name of the log file to a generic `replication.log`?  Using the name of a particular implementation in the log filename will often become a hassle down the road, like if the replication system changes.  replication.log would work for anything.

userguide/storage.rst
```diff
@@ -2014,10 +2013,10 @@ Replication schedules are typically paired with
 :ref:`Periodic Snapshot Tasks` to generate local copies of important
 data and replicate these copies to a remote system.
 
-Replications require a source system with datset snapshots and a
+Replications require a source system with dataset snapshots and a
 destination that can store the copied data. Remote replications require
-a saved :ref:`SSH Connection <SSH Connections>` on the source system and
-the destination system must be configured to allow :ref:`SSH`
+a saved :ref:`SSH Connection <SSH Connections>` on the local system and
```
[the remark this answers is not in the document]
themylogin: Problem is that we now have push replication, so "local = source" and "remote = destination" does not work anymore, that's why I am suggesting this.

userguide/storage.rst
```diff
@@ -2091,37 +2090,35 @@ different :guilabel:`Transport` options:
...
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | Netcat Active Side        | NCT       | drop-down menu | Choose the system that will use `nc(1) <https://www.freebsd.org/cgi/man.cgi?query=nc>`__ to open TCP ports and  |
-   |                           |           |                | configure the connection between the two systems. *LOCAL* designates the system that is creating the            |
-   |                           |           |                | replication as the active side of the connection. *REMOTE* designates the replication destination as the active |
-   |                           |           |                | side.                                                                                                           |
+   | Netcat Active Side        | NCT       | drop-down menu | Choose the system (*LOCAL* or *REMOTE*) that will open TCP ports to configure the connection between the two    |
```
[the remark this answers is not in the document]
themylogin: To create a TCP connection one side (server) needs to open TCP port to which other side (client) can connect to.

userguide/storage.rst
```diff
@@ -2167,18 +2164,18 @@ different :guilabel:`Transport` options:
...
    | Matching Schedule         | LOC       |                | the :guilabel:`Schedule` values to the :guilabel:`Snapshot Replication Schedule`.                               |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | Replicate from scratch if | SSH, NCT, | checkbox       | Synchronizes source and destination snapshots. When a source snapshot is determined to be out of sync with the  |
-   | incremental is not        | LOC       |                | destination system, destroy the related destination snapshot and upload a full copy of the source snapshot.     |
-   | possible                  |           |                |                                                                                                                 |
+   | Replicate from scratch if | SSH, NCT, | checkbox       | If this is enabled and destination system does have snapshots but none of them matches any snapshot on source   |
```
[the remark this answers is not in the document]
themylogin: I like "Perform full replication if incremental is impossible" and our UI can tolerate this perfectly. Can't think of anything that can be significantly shorter and still explain a purpose of this clearly. Do you think we should make this change?

userguide/storage.rst
themylogin: User won't need to interact with this file, we have logs in UI that are split per-task. I've named it after zettarepl because that's how other files in /var/log are named: nginx/access.log, not webserver/access.log; proftpd.log, not ftp.log, etc. If the replication system changes, replication.log will change it's format too and being on the same place won't give nobody any help.

userguide/storage.rst
```diff
@@ -1948,9 +1948,9 @@ summarizes the fields in this screen.
...
    +-------------------+-------------------+--------------------------------------------------------------------------------------------+
    | Exclude           | string            | Exclude specific child datasets from the snapshot. Use with :guilabel:`Recursive`          |
-   |                   |                   | snapshots. List paths to any child datasets to exclude. Separate multiple entries with a   |
-   |                   |                   | comma (:literal:`,`). Example: :samp:`pool1/dataset1/child1`. A recursive snapshot of      |
-   |                   |                   | :file:`pool1/dataset1` includes all child datasets except :file:`child1`.                  |
+   |                   |                   | snapshots. List names of child datasets to exclude, one per line.                          |
+   |                   |                   | Example: :samp:`pool1/dataset1/child1`. A recursive snapshot of :file:`pool1/dataset1`     |
+   |                   |                   | includes all child datasets except :file:`child1` and it's descendants.                    |
```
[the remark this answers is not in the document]
themylogin: Do we (or anybody other) say "dataset tank/data is below dataset tank"?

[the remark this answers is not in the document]
wblock: How about "includes all child datasets except child1 and any datasets inside it."?

userguide/storage.rst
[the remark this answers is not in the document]
wblock: Okay, so it's really "Choose the system that will open receiving ports." Not clear how that is the "active" side, maybe call it the "listening side"?

userguide/storage.rst
```diff
@@ -2014,12 +2013,11 @@ Replication schedules are typically paired with
...
 data and replicate these copies to a remote system.
 
-Replications require a source system with datset snapshots and a
+Replications require a source system with dataset snapshots and a
 destination that can store the copied data. Remote replications require
 a saved :ref:`SSH Connection <SSH Connections>` on the source system and
 the destination system must be configured to allow :ref:`SSH`
-connections. Replications to the local system do not require configuring
-SSH.
+connections. Local replications do not require configuring SSH.
```
wblock: Clarify "require configuring":
```suggestion
connections. Local replications do not use SSH.
```

userguide/storage.rst
```diff
@@ -2063,7 +2061,7 @@ Creating remote, local, %brand% legacy engine, and one-time replications
 are all possible with this screen.
 
 The :guilabel:`Transport` value changes many of the options for
-replication. :ref:`Table %s <zfs_add_replication_task_opts_tab>` uses
+replication. :numref:`Table %s <zfs_add_replication_task_opts_tab>` uses
```
```suggestion
replication. :numref:`Table %s <zfs_add_replication_task_opts_tab>` shows
```

userguide/storage.rst
```diff
@@ -2063,7 +2061,7 @@ Creating remote, local, %brand% legacy engine, and one-time replications
...
 shortened versions of the :guilabel:`Transport` methods in the
```
```suggestion
abbreviated names of the transport methods in the
```

userguide/storage.rst
```diff
@@ -2063,7 +2061,7 @@ Creating remote, local, %brand% legacy engine, and one-time replications
...
 :literal:`Transport` column to show which fields appear with the
```
```suggestion
:literal:`Transport` column to identify fields which appear when that method is selected.
```

userguide/storage.rst
```diff
@@ -2091,37 +2089,35 @@ different :guilabel:`Transport` options:
    +===========================+===========+================+=================================================================================================================+
    | Name                      | ALL       | string         | Enter a descriptive :guilabel:`Name` for the replication.                                                       |
```
wblock: Use upper/lower "All" to differentiate the non-literal "All" from the specific types in upper case.

userguide/storage.rst
```diff
@@ -2130,15 +2126,15 @@ different :guilabel:`Transport` options:
...
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
    | Exclude Child Datasets    | SSH, NCT, | string         | Exclude specific child dataset snapshots from the replication. Use with :guilabel:`Recursive` snapshots. List   |
-   |                           | LOC       |                | paths to any child datasets to exclude. Example: :samp:`pool1/dataset1/child1`. A recursive replication of      |
-   |                           |           |                | :file:`pool1/dataset1` snapshots will include all child dataset snapshots except :file:`child1`.                |
+   |                           | LOC       |                | child dataset names to exclude. Example: :samp:`pool1/dataset1/child1`. A recursive replication of              |
+   |                           |           |                | :file:`pool1/dataset1` snapshots includes all child dataset snapshots except :file:`child1`.                    |
```
```suggestion
   |                           |           |                | :file:`pool1/dataset1` snapshots includes all child datasets except :file:`child1`.                    |
```

userguide/storage.rst
```diff
@@ -2130,15 +2126,15 @@ different :guilabel:`Transport` options:
...
-   |                           |           |                | the name of the dataset that will store replicated snapshots.                                                   |
+   |                           |           |                | the name of the dataset to store replicated snapshots.                                                          |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
    | Recursive                 | ALL       | checkbox       | Replicate all child dataset snapshots. Set to make :guilabel:`Exclude Child Datasets` visible.                  |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
    | Exclude Child Datasets    | SSH, NCT, | string         | Exclude specific child dataset snapshots from the replication. Use with :guilabel:`Recursive` snapshots. List   |
```
```suggestion
   | Exclude Child Datasets    | SSH, NCT, | string         | Exclude specific child datasets from the replication. Use with :guilabel:`Recursive` snapshots. List   |
```

userguide/storage.rst
```diff
@@ -2167,18 +2163,18 @@ different :guilabel:`Transport` options:
...
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | Begin                     | SSH, NCT, | drop-down menu | Define a starting time when the replication cannot run. A replication that is in progress can continue to run   |
-   |                           | LOC       |                | past this time.                                                                                                 |
-   +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
-   | End                       | SSH, NCT, | drop-down menu | Define an ending time for the restriction on activating replication schedules.                                  |
+   | Begin                     | SSH, NCT, | drop-down menu | Set a starting time for the replication task to run.                                                            |
```
```suggestion
   | Begin                     | SSH, NCT, | drop-down menu | Start time for the replication task.                                                            |
```

userguide/storage.rst
```diff
@@ -2167,18 +2163,18 @@ different :guilabel:`Transport` options:
...
    |                           | LOC       |                |                                                                                                                 |
    +---------------------------+-----------+----------------+-----------------------------------------------------------------------------------------------------------------+
+   | End                       | SSH, NCT, | drop-down menu | Set an ending time for the replication task. A replication that is in progress can continue to run past this    |
```
```suggestion
   | End                       | SSH, NCT, | drop-down menu | End time for the replication task. A replication that is already running can continue past this    |
```

PR #1304, opened 2019-07-11
NAS-101349 UPS events now use Alerts

a-stjohn:

PR #1305, opened 2019-07-11
NAS-101349 UPS events now use Alerts

a-stjohn:
Cherry picked PR #1304

### DIFF
--- a/userguide/snippets/alertevents.rst
+++ b/userguide/snippets/alertevents.rst
@@ -109,7 +109,7 @@ Some of the conditions that trigger an alert include:
 
 * a connected Uninterruptible Power Supply (UPS) switches to battery
   power, switches to line power, communication with the UPS is lost or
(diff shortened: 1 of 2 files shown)


userguide/snippets/alertevents.rst
```diff
@@ -109,6 +109,10 @@ Some of the conditions that trigger an alert include:
 
 * a :ref:`Cloud Sync` task fails
 
+* a connected Uninterruptible Power Supply (UPS) switches to battery
+  power, switches to line power, communication with the UPS is lost or
+  establisted, the battery is low, or the battery needs to be replaced.
```
wblock: These are not sentences, so no period.  Well, they are sentences, but the style of this section has been as short non-sentences.
```suggestion
  establisted, the battery is low, or the battery needs to be replaced
```

PR #1309, opened 2019-07-15
NAS-100329 Doc VMs individual log files

a-stjohn:


userguide/intro.rst
```diff
@@ -270,6 +270,9 @@ These screen options have changed:
 * The names *global*, *homes*, and *printers* cannot be used in
   :menuselection:`Sharing --> Windows (SMB) --> Add Windows (SMB) Share`.
 
+* Log files are now generated for each VM and located at
+  :file:`/var/log/vm`.
```
wblock: Not clear whether /var/log/vm is a directory or "vm" is replaced with the VM name.

userguide/vms.rst
```diff
@@ -405,6 +405,9 @@ When a VM is already running, these buttons are available:
   VM must have a VNC device and :guilabel:`VNC Web` enabled in that
   device.
 
+Each VM generates a specific log file. The log files for VMs are in
```
wblock: Same as above.  I'd guess each VM has a subdirectory in /var/logs/vm/, but this does not make that clear.

wblock:
Note that I removed the word "created" from both entries, since it didn't mean anything in that context.

PR #1311, opened 2019-07-15
NAS-102098 doc new column in Tasks->Replication Tasks

a-stjohn:


userguide/tasks.rst
```diff
@@ -959,6 +959,13 @@ Click :guilabel:`NEXT` to see the replication configuration summary.
 :guilabel:`BACK` returns to the previous screens to adjust the
 replication settings.
 
+Created replication tasks are displayed in
+:menuselection:`Tasks --> Replication Tasks`. The default columns
```
wblock: We should not list the default columns anywhere, because that will be out of sync as soon as the user changes them or the defaults change.  If we don't think it will be shown, we should just tell them to enable it.  Or just do that by default.
```
The *Last Snapshot* column shows the date and time of the last
snapshot taken for a given replication task.
If the :guilabel:`Last Snapshot` column is not shown, it can be
enabled from the :guilabel:`COLUMNS` button.
```

userguide/intro.rst
```diff
@@ -229,6 +229,9 @@ These screen options have changed:
 * The :guilabel:`Name` field has been added to
   :menuselection:`Tasks --> Replication Tasks --> Add`.
 
+* The *Last Snapshot* column has been added to
```
```suggestion
* A :guilabel:`Last Snapshot` column has been added to
```

PR #1323, opened 2019-07-18
NAS-100280: Update Alert Services docs

Mrt134:
- Add intro entry for new and removed fields.
- Update text in system/alert services section.
- Update image related to text.
- HTML build test: no issues.


userguide/system.rst
```diff
@@ -1165,14 +1165,16 @@ Click |ui-add| to display the :guilabel:`Add Alert Service` form,
 
 
 Select the :guilabel:`Type` to choose an alert service to configure.
+
+Each alert service can also be associated with a specific alert severity
```
wblock: This is not very clear, and would be helped by an example.
```
Alert services can be set for a particular :guilabel:`Level` of severity.
All alerts of that level are then sent out with that alert service. For
example, if the *E-Mail* alert service :guilabel:`Level` is set to *Info*,
any Info level alerts are sent by that service. Multiple alert services
can be set to the same level. For instance, *Critical* alerts could be
sent both by email and PagerDuty by setting both of those alert
services to that level.
```

PR #1326, opened 2019-07-19
intro entry for removal of docker

a-stjohn:
Intro entry for PR #1194


userguide/intro.rst
```diff
@@ -159,7 +159,7 @@ This software has been added or updated:
 
 * Docker has been removed as a
   :menuselection:`Virtual Machines`
-  option.
+  option and the Grub boot loader is now supported.
```
wblock: It's not clear how these two things are related.

a-stjohn: When creating a docker VM, when it existed, it would use the Grub boot loader to create the Docker VM. Now Grub is a stand alone thing for VMs.

wblock: So this should be two intro entries?

a-stjohn: done

PR #1338, opened 2019-07-23
NAS-100308 doc engineering mode in web UI

a-stjohn:


wblock:
"Engineering mode" was decided against, so this was overtaken by events.

PR #1340, opened 2019-07-24
Intro entry for viewing replication tasks logs

a-stjohn:


userguide/intro.rst
```diff
@@ -232,6 +232,9 @@ These screen options have changed:
 * The :guilabel:`Hold Pending Snapshots` checkbox has been added to
   :menuselection:`Tasks --> Replication Tasks --> ADD`.
 
+* Replication Task log files can be easily displayed and downloaded in
```
wblock: Leave out value judgements.
```suggestion
* Replication Task log files can be displayed and downloaded in
```

userguide/intro.rst
```diff
@@ -232,6 +232,9 @@ These screen options have changed:
...
+  :menuselection:`Tasks --> Replication Tasks`.
```
wblock: Use a :ref: instead of menuselection.

userguide/intro.rst
```diff
@@ -247,8 +247,8 @@ These screen options have changed:
 * The :guilabel:`Hold Pending Snapshots` checkbox has been added to
   :menuselection:`Tasks --> Replication Tasks --> ADD`.
 
-* Replication Task log files can be displayed and downloaded in
-  :menuselection:`Tasks --> Replication Tasks`.
+* Replication Task log files are displayed and downloaded in
```
wblock: Displaying or downloading log files is an optional action, so "can be" was correct here.

PR #1359, opened 2019-08-02
NAS-102136 doc additional disk info being added to disk page

a-stjohn:


userguide/intro.rst
```diff
@@ -325,6 +325,9 @@ These screen options have changed:
   :menuselection:`Storage -->` |ui-options| :menuselection:`--> Delete Dataset`
   when trying to delete a dataset that is busy.
 
+* Disk type, rotation rate, and manufacture information can be viewed
```
```suggestion
* Disk type, rotation rate, and manufacturer information can be viewed
```

userguide/storage.rst
```diff
@@ -1772,7 +1772,8 @@ management settings, acoustic level settings, and whether
 :ref:`S.M.A.R.T.` tests are enabled. The pool associated with the disk
 is displayed in the :guilabel:`Pool` column. *Unused* is displayed if
 the disk is not being used in a pool. Click :guilabel:`COLUMNS` to
-adjust the table.
+add or remove additional information as columns. Any information that
```
wblock: This is kind of oddly worded.  The important part here is that the COLUMNS menu is for adding or removing columns shown by the table.  I'm not clear on what the last sentence is saying.

userguide/intro.rst
```diff
@@ -303,6 +303,13 @@ These screen options have changed:
 * :ref:`Directory Services Monitor <Directory Services>` has been added
   to the top toolbar row.
 
+* Disk type, rotation rate, and manufacturer information can be viewed
+  on the :ref:`Disks` page.
+
+* The :guilabel:`Disable LDAP user/group cache` checkbox has been added
```
wblock: This should be title case, probably in the UI also.
```suggestion
* The :guilabel:`Disable LDAP User/Group Cache` checkbox has been added
```

userguide/storage.rst
```diff
@@ -1765,8 +1765,10 @@ device name, serial number, size, advanced power
 management settings, acoustic level settings, and whether
 :ref:`S.M.A.R.T.` tests are enabled. The pool associated with the disk
 is displayed in the :guilabel:`Pool` column. *Unused* is displayed if
-the disk is not being used in a pool. Click :guilabel:`COLUMNS` to
-adjust the table.
+the disk is not being used in a pool. Click :guilabel:`COLUMNS` and
```
wblock: I don't know that we need to keep saying this.  If we do (and let's decide that), it can be a snippet.  But I think it should be mentioned once and then ignored.  Maybe a "how to use the UI" section at the beginning of the guide.

userguide/intro.rst
a-stjohn: @wblock webUI PR: https://github.com/freenas/webui/pull/2648

PR #1364, opened 2019-08-05
NAS-100007: Add entry for AD related alerts

Mrt134:
- Investigated directoryservices.rst and found text related to the alerts was already present.
- As previously discussed, no intro entry is necessary.
- HTML build test: no issues.

### DIFF
--- a/userguide/snippets/alertevents.rst
+++ b/userguide/snippets/alertevents.rst
@@ -93,7 +93,7 @@ Some of the conditions that trigger an alert include:
 
 * hardware events detected by an attached :ref:`IPMI` controller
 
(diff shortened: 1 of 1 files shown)


userguide/snippets/alertevents.rst
```diff
@@ -93,6 +93,8 @@ Some of the conditions that trigger an alert include:
 
 * hardware events detected by an attached :ref:`IPMI` controller
 
+* error found with the :ref:`Active Directory` connection
```
a-stjohn:
```suggestion
* an error found with the :ref:`Active Directory` connection
```

PR #1374, opened 2019-08-08
NAS-101567 doc location of ssh keys.

a-stjohn:


userguide/system.rst
```diff
@@ -220,7 +220,9 @@ This screen also contains these buttons:
   :menuselection:`System --> System Dataset`.
 
   .. note:: :ref:`SSH` keys are not stored in the configuration database
-     and must be backed up separately.
+     and must be backed up separately. System host keys are stored in
+     :file:`/usr/local/etc/ssh/ssh_hosts*`. The root user keys are
```
wblock: It's not clear what the star means, but probably pattern matching.  Be explicit about these, and I suspect they actually are ssh_host_ (no "s").  Maybe say "System host keys are files with names beginning with `ssh_host_` in :file:`/usr/local/etc/ssh/`.

PR #1376, opened 2019-08-08
NAS-102104 doc SMB recycle bin caveat.

a-stjohn:
Will need an 11.2-angular and 11.2-legacy commit once text is approved.


userguide/sharing.rst
```diff
@@ -920,7 +920,9 @@ provides more details for each configurable option.
    |                                |               |          |                                                                                                                                                      |
    +--------------------------------+---------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------+
    | Export Recycle Bin             | checkbox      | ✓        | Set for deleted files to move to :file:`.recycle` in the root folder of the share. The :file:`.recycle` directory can be deleted to reclaim space    |
-   |                                |               |          | and is recreated whenever a file is deleted.                                                                                                         |
+   |                                |               |          | and is recreated whenever a file is deleted. If a file being deleted is in a subdirectory and larger than 20 MiB, adjust the                         |
```
wblock: Not clear what the parameter does.  I suspect that files larger than the set size are not kept in .recycle, but just deleted.  Please clarify.

userguide/sharing.rst
```diff
@@ -920,7 +920,10 @@ provides more details for each configurable option.
    |                                |               |          |                                                                                                                                                      |
    +--------------------------------+---------------+----------+------------------------------------------------------------------------------------------------------------------------------------------------------+
    | Export Recycle Bin             | checkbox      | ✓        | Set for deleted files to move to :file:`.recycle` in the root folder of the share. The :file:`.recycle` directory can be deleted to reclaim space    |
-   |                                |               |          | and is recreated whenever a file is deleted.                                                                                                         |
+   |                                |               |          | and is recreated whenever a file is deleted. If a file being deleted is in a subdirectory and larger than 20 MiB, the file is deleted permanently.   |
```
wblock: I don't understand the part about being in a subdirectory.  I think the setting is just to keep huge files from being stored in the recycle bin.  So if a file is larger than that limit, it is not moved but just immediately deleted.

userguide/sharing.rst
```diff
@@ -920,7 +920,10 @@ provides more details for each configurable option.
...
+   |                                |               |          | To move files larger than 20 MiB into :file:`.recycle`, adjust the :guilabel:`Auxiliary Parameter` :command:`crossrename:sizelimit=` to a            |
```
wblock: I'd explain this as "To allow larger files to be moved to the Recycle Bin rather than deleted immediately, adjust the ... to a larger size.  For example..."

userguide/sharing.rst
anodos325: No. If the file is in a child dataset, the size of the file being recycled is an issue. When the file being recycled is located within the same dataset, the file is simply moved to the recycle bin via rename. There is no size limit when the file is deleted inside the same dataset where the recycle bin is located.

wblock: Ah, that makes sense. Attempted summary:
```
Files that are deleted from the same dataset are moved to the Recycle Bin
and do not take any additional space. When the files are in a different dataset
or a child dataset, they are copied to the dataset where the Recycle Bin is
located. To prevent excessive space usage, files larger than 20 MiB are deleted
rather than moved. Adjust the :guilabel:`Auxiliary Parameter`
:samp:`crossrename:sizelimit=` setting to allow larger files. For example,
:samp:`crossrename:sizelimit={50}`allows moves of files up to 50 MiB in size.
```

PR #1378, opened 2019-08-08
NAS-101049 intro entry for grub boot loader (again)

a-stjohn:


userguide/intro.rst
```diff
@@ -360,6 +360,8 @@ These screen options have changed:
 
 * Docker has been removed as a :ref:`Virtual Machines <VMs>` option.
 
+* Grub boot loader is supported for virtual machines.
```
```suggestion
* Grub boot loader support has been added for virtual machines that will not boot with other loaders.
```

PR #1315, opened 2019-07-17
NAS-100543 doc force delete dataset when busy dialog.

a-stjohn:

PR #1391, opened 2019-08-13
NAS-100543 clean up docs for deleting dataset

a-stjohn:
clean up docs with regard to [dru's comment](https://jira.ixsystems.com/browse/NAS-100543?focusedCommentId=97477&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-97477) on PR #1315


userguide/storage.rst
```diff
@@ -1026,7 +1026,8 @@ any objects stored within the dataset. To remove the dataset, set
 :guilabel:`Confirm`, click :guilabel:`DELETE DATASET`, verify
 that the correct dataset to be deleted has been chosen by entering the
 dataset name, and click :guilabel:`DELETE`. When the dataset is busy, a
-force delete dialog option appears, showing what has the dataset in use.
```
wblock: I think the complaint here was that it didn't tell what a dataset being "busy" meant.  I suggest (verify/adjust the wording):
`When the dataset has active shares or is still being used by other parts of the system, the dialog shows  what is still using it and allows forcing the deletion anyway. Caution: forcing the deletion of an in-use dataset can cause data loss or other problems.`

PR #1418, opened 2019-08-15
NAS-100372: Add text about automatic swap partitions

Mrt134:
- Add text about behavior and restrictions in install.rst
- Add note to intro changelog
- HTML build test: no issues.

### DIFF
--- a/userguide/install.rst
+++ b/userguide/install.rst
@@ -300,9 +300,11 @@ Use the arrow keys to highlight the destination SSD, hard drive,
 |usb-stick|, or virtual disk. Press the :kbd:`spacebar` to select
 it.
 
-To mirror the |os-device|, move to the second device and press
-:kbd:`spacebar` to select it also. Selecting a non-USB device larger
-than 64 GiB also creates a 16 GiB swap partition.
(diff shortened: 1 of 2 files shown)


userguide/install.rst
```diff
@@ -298,9 +298,14 @@ system.
 
 Use the arrow keys to highlight the destination SSD, hard drive,
 |usb-stick|, or virtual disk. Press the :kbd:`spacebar` to select
-it. To mirror the |os-device|, move to the second device and press
-:kbd:`spacebar` to select it also. After making these selections,
-press :kbd:`Enter`. The warning shown in
+it.
+
+To mirror the |os-device|, move to the second device and press
+:kbd:`spacebar` to select it also. 16 GiB swap partitions are created
```
wblock: It's not clear what the second sentence is saying.  I think it means:
"If the device is not a USB device, and larger than 64 GiB, a 16 GiB swap partition is also created on it."

userguide/intro.rst
```diff
@@ -151,6 +151,10 @@ These major features are new in this version:
 
 This software has been added or updated:
 
+* The %brand% installer creates 16 GiB swap partitions when installing
```
wblock: Reorder as above, if correct.

PR #1430, opened 2019-08-19
NAS-102865 rewrite "Replacing disks to grow a pool" section

a-stjohn:


userguide/storage.rst
```diff
@@ -2044,18 +2044,19 @@ be impacted by their removal.
...
-
-However, this is not an option if there are no open drive ports and a
-SAS/SATA HBA card cannot be added. In this case, one disk at a time
-can be replaced with a larger disk, waiting for the resilvering
-process to incorporate the new disk into the pool, then repeating with
-another disk until all of the original disks have been replaced.
-
-The safest way to replace a drive is to use a spare drive port or an
-eSATA port and a hard drive dock. The process is:
+Increasing the size of a :ref:`pool <Pools>` can be an easy task with
```
wblock: This is not quite correct.  Hot-swap trays just make it possible to replace drives without shutting the system off.  It does not remove the need to select the options from the menus.  In other words, it does not replace the steps, it just makes the "shut down the system" step not necessary if you have hot-swap.

PR #1442, opened 2019-08-20
NAS-101229: Add text about Cron job command length limit.

Mrt134:
- HTML build test: no issues.


userguide/tasks.rst
```diff
@@ -81,7 +81,7 @@ lists the configurable options for a cron job.
    |                     |                             |                                                                                                         |
    +---------------------+-----------------------------+---------------------------------------------------------------------------------------------------------+
    | Command             | drop-down menu              | Enter the **full path** to the command or script to be run. If it is a script, testing it at the        |
-   |                     |                             | command line is recommended to ensure it works.                                                         |
+   |                     |                             | command line is recommended to ensure it works. Strings with more than 1000 characters are not allowed. |
```
wblock: "More than x characters" is oddly stated.  Flipping the logic makes it clearer.  (Table space is broken, please test.)
```suggestion
   |                     |                             | command line first is recommended. Strings up to 1000 characters are allowed. |
```

PR #1444, opened 2019-08-20
NAS-101297: Update manual snapshot and vmware-snapshots sections

Mrt134:
- Add text about possible checkboxes to manual snapshot dialog
- Add text about field behaviors in VMware-Snapshots section
- HTML build test: no issues.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1024,9 +1024,10 @@ the origin filesystem becomes a clone of the clone making it possible
 to destroy the filesystem that the clone was created from. Otherwise,
 a clone cannot be deleted while the origin filesystem exists.
 
-**Create Snapshot:** create a one-time snapshot. To schedule the
-regular creation of snapshots, instead use
-:ref:`Periodic Snapshot Tasks`.
+**Create Snapshot:** create a one-time snapshot. A dialog opens to name
+the snapshot. Options to include child datasets in the snapshot and
+synchronize with VMware can also be shown. To schedule snapshot
+creation, use :ref:`Periodic Snapshot Tasks`.
 
 
 #ifdef freenas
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -1718,28 +1719,27 @@ summarizes the available options.
...
-   |                |                             | :guilabel:`FETCH DATASTORES` to populate the menu, then select the datastore to be synchronized.            |
-   |                |                             |                                                                                                             |
-   +----------------+-----------------------------+-------------------------------------------------------------------------------------------------------------+
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+   | Setting        | Value          | Description                                                                                                 |
+   +================+================+=============================================================================================================+
+   | Hostname       | string         | Enter the IP address or hostname of the VMware host. When clustering, use the IP of the vCenter server for  |
```
a-stjohn: Suggestion:
```
Use the IP of the vCenter server when clustering.
```

userguide/storage.rst
```diff
@@ -1718,28 +1719,27 @@ summarizes the available options.
...
+   |                |                | the cluster.                                                                                                |
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+   | Username       | string         | Enter the username on the VMware host with permission to snapshot virtual machines.                         |
```
a-stjohn: I'm kinda confused with that wording. If I understand correctly, does my suggestion make more sense??

```suggestion
   | Username       | string         | Enter the username of the VMware host with permission to snapshot virtual machines.                         |
```

userguide/storage.rst
```diff
@@ -1718,28 +1719,27 @@ summarizes the available options.
...
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+   | Password       | string         | Enter the password associated with :guilabel:`Username`.                                                    |
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+   | ZFS Filesystem | browse button  | :guilabel:`Browse` to the filesystem to snapshot.                                                           |
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+   | Datastore      | drop-down menu | After entering the :guilabel:`Hostname`, :guilabel:`Username`, and :guilabel:`Password`, click              |
+   |                |                | :guilabel:`FETCH DATASTORES` to populate the menu, then select the datastore to be synchronized.            |
```
a-stjohn: Remove pause.

```suggestion
   |                |                | :guilabel:`FETCH DATASTORES` to populate the menu and select the datastore to be synchronized.              |
```

userguide/storage.rst
```diff
@@ -1718,28 +1719,27 @@ summarizes the available options.
...
+   +----------------+----------------+-------------------------------------------------------------------------------------------------------------+
+
+
+%brand% attempts to connect to the VMware host after the credentials are
+entered and populate the :guilabel:`ZFS Filesystem` and
+:guilabel:`Datastore` drop-down menus. Choosing a datastore also selects
```
a-stjohn: I read the first sentence several times before understanding. Suggestion: make 2 sentences-
```
%brand% attempts to connect to the VMware host after the credentials
are entered. It also populates the :guilabel:`ZFS Filesystem` and
:guilabel:`Datastore` drop-down menus.
```
or something like that...

userguide/storage.rst
wblock: I think it's clearer the first way (because it only applies when clustering), but s/IP/IP address or hostname/.

userguide/storage.rst
wblock: Agreed, it's not clear.  Is it the user or the host that has permission?

userguide/storage.rst
wblock: It's two actions, populate the menu, then select the datastore.  So the comma should be there.

userguide/storage.rst
wblock: Yes, break the two thoughts into two sentences.  Also, "attempts" implies that it often fails, which is probably not true.
`After the credentials are entered, %brand% connects to the VMware host and populates the ...`

PR #1446, opened 2019-08-20
NAS-101416: Update Network/Interfaces Parent Interface description

Mrt134:
- Add note that bridges cannot be selected
- HTML build test: no issues


userguide/network.rst
```diff
@@ -237,7 +237,8 @@ which settings are available with each interface type.
    |                     |                | Aggregation | interfaces have manually assigned IP addresses.                                                           |
    +---------------------+----------------+-------------+-----------------------------------------------------------------------------------------------------------+
    | Parent Interface    | drop-down menu | VLAN        | Select the VLAN Parent Interface. Usually an Ethernet card connected to a switch port configured for      |
-   |                     |                |             | the VLAN. New link aggregations are not available until the system is restarted.                          |
+   |                     |                |             | the VLAN. A *bridge* cannot be selected as a parent interface. New link aggregations are not available    |
```
a-stjohn: Do we have a "Link Aggregations" section? Might be nice to have "link aggregations" as a :ref: to that section if we hve one.

```suggestion
   |                     |                |             | the VLAN. A *bridge* cannot be selected as a parent interface. New link aggregations are not available    |
```

wblock: We do:
`` :ref:`Link Aggregations` ``

PR #1469, opened 2019-08-22
NAS-102881 rework help text docs. doc'ed dragability of boxes

a-stjohn:


userguide/intro.rst
```diff
@@ -432,6 +432,9 @@ These screen options have changed:
 
 * Right-click help dialog has been added to the :ref:`Shell`.
 
+* :ref:`Help text <help text>` boxes can be pinned and moved on the
```
wblock: I'm not really sure what pinning does now.  Maybe the last part should be "... pinned in place."?
```suggestion
* :ref:`Help text <help text>` boxes can be moved or pinned.
```

userguide/intro.rst
```diff
@@ -432,6 +432,9 @@ These screen options have changed:
...
+  screen.
```
```suggestion
```

PR #1508, opened 2019-09-04
NAS-102997 doc NAT and BPF being mutually exclusive.

a-stjohn:
I don't think this needs an intro entry.

### DIFF
--- a/userguide/jails.rst
+++ b/userguide/jails.rst
@@ -255,7 +255,8 @@ a new jail.
    |                           |                   |                                                                                                         |
    +---------------------------+-------------------+---------------------------------------------------------------------------------------------------------+
    | NAT                       | checkbox          | Network Address Translation (NAT). When set, the jail is given an internal IP address and               |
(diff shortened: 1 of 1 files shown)


userguide/jails.rst
```diff
@@ -267,7 +268,7 @@ a new jail.
    | Berkeley Packet Filter    | checkbox          | Use the Berkeley Packet Filter to data link layers in a protocol independent fashion. Unset by default  |
    |                           |                   | to avoid security vulnerabilities. See                                                                  |
    |                           |                   | `BPF(4) <https://www.freebsd.org/cgi/man.cgi?query=bpf>`__                                              |
-   |                           |                   | for more details.                                                                                       |
+   |                           |                   | for more details. Cannot be set if :guilabel:`NAT` is set.                                              |
```
Mrt134:
```suggestion
   |                           |                   | for more details. Cannot be set when :guilabel:`NAT` is set.                                            |
```

PR #1510, opened 2019-09-05
NAS-102850: Update SSH Keypairs text to indicate encrypted/passphrase…

Mrt134:
… keypairs aren't supported.

- HTML build test: no issues.


userguide/system.rst
```diff
@@ -1702,7 +1702,8 @@ SSH Keypairs
 SSH public and private keypairs in
 :menuselection:`System --> SSH Keypairs`.
 These are generally used when configuring :ref:`SSH Connections` or
-*SFTP* :ref:`Cloud Credentials`.
+*SFTP* :ref:`Cloud Credentials`. Encrypted keypairs or keypairs with
+passphrases are not supported.
```
a-stjohn: Can you have a non-encrypted keypair with a passphrase? If so, then LGTM. Otherwise, I would say it could be reduced to
```
Encrypted keypairs are not supported.
```

Mrt134: hm, good question - I'll look into it a bit more before merging.

Mrt134: Looks like passphrase is only used in conjunction with encryption: https://www.ssh.com/ssh/passphrase - will rework text.

wblock: Passphrases imply encryption.  But the reader doesn't necessarily know that, which was why this was worded to include both.  Please go back to that wording.

PR #1514, opened 2019-09-05
NAS-102450: Update 'WebGUI HTTP -> HTTPS Redirect' description.

Mrt134:
- Small cleanup
- Add note about HSTS max age value.
- Not sure if this is needed because it is an implementation detail, but will review and confirm at a later date.
- HTML build test: no issues.


userguide/system.rst
```diff
@@ -168,11 +168,8 @@ settings.
...
-   | WebGUI HTTP ->       | checkbox       | Set to redirect *HTTP* connections to *HTTPS*.                                                                           |
-   | HTTPS Redirect       |                | A :guilabel:`GUI SSL Certificate` is required for *HTTPS*.                                                               |
-   |                      |                |                                                                                                                          |
-   |                      |                |                                                                                                                          |
-   |                      |                |                                                                                                                          |
+   | WebGUI HTTP ->       | checkbox       | Redirect *HTTP* connections to *HTTPS*. A :guilabel:`GUI SSL Certificate` is required for *HTTPS*. Activating this also  |
+   | HTTPS Redirect       |                | sets the HTTP Strict Transport Security (HSTS) maximum age to :literal:`31536000`.                                       |
```
wblock: Can we get a link that defines what HSTS is and maybe why that magic value is used?

PR #1537, opened 2019-09-16
NAS-103089: Document 'Region' S3 Cloud Cred. field

Mrt134:
- Synchronize with https://github.com/freenas/webui/pull/2766 when it is approved and merged.
- HTML build test: no issues.

### DIFF
--- a/userguide/intro.rst
+++ b/userguide/intro.rst
@@ -219,8 +219,8 @@ These screen options have changed:
 * :guilabel:`From Name` has been added to :ref:`Email`.
 
 * :guilabel:`Periodic Notification User` has been removed from the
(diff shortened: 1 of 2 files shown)


themylogin:
@Mrt134 if you don't want to mention AWS GovCloud in tooltip, maybe we should mention it at least in the doc? Because that looks like the unique case (yet) when this is necessary.

Mrt134:
I will look into this and update accordingly

Mrt134:
Ok, solution is to add links and examples of an AWS GovCloud configuration to the UI help text and keep the text synchronized with the guide. I'll update this PR when the final help text is decided in webui PR 2766.

PR #1542, opened 2019-09-17
testing heading perma-links

a-stjohn:


a-stjohn:
Closed because it doesn't seem that there's an easy way to insert a permalink with a blank header.

PR #1548, opened 2019-09-19
NAS-102368: Add 'Default ACL Options' to "ACL Options" table

Mrt134:
- Update related image
- Rework the help text description. Open a webui PR with any updates to sync back to the help text.
- HTML build test: no issues.


Mrt134:
I don't think so. The ACL manager is completely new for 11.3 and there is already an intro entry for the new page.

PR #1554, opened 2019-09-20
NAS-100995: Unlock SED before including in pool

Mrt134:
- Proposing a small change to SED description. Not sure if a docs change is really needed for this ticket.
- HTML build test: no issues.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -790,7 +790,8 @@ command line.
 By default, SEDs are not locked until the administrator takes ownership
 of them. Ownership is taken by explicitly configuring a global or
 per-device password in the %brand% |web-ui| and adding the password to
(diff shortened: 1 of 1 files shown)


userguide/system.rst
```diff
@@ -790,7 +790,7 @@ command line.
 By default, SEDs are not locked until the administrator takes ownership
 of them. Ownership is taken by explicitly configuring a global or
 per-device password in the %brand% |web-ui| and adding the password to
-the SEDs.
+the SEDs. This also allows %brand% to create :ref:`pools` with SEDs.
```
wblock: Not clear what _This_ refers to here.  The added sentence is kind of specific, implying that other things might be possible without it.  I think that what this is really trying to say is that when ownership/passwords for SEDs are entered into %brand%, it can use them without manual unlocking.

PR #1569, opened 2019-09-27
NAS-103265: Remove iocage cli instructions.

Mrt134:
- Port a small change to the Jail intro section.
- Update the Jail Software section intro to show the process from the UI.
- Add new conf.py replacement for the Jail Shell icon.
- Replace text example with screenshot.
- HTMl build test: no issues.


userguide/conf.py
```diff
@@ -116,6 +116,7 @@
 .. |ui-jail-start|        replace::  (Start)
 .. |ui-jail-stop|         replace::  (Stop)
 .. |ui-jail-update|       replace::  (Update)
+.. |ui-jail-shell|        replace:: 塞 SHELL
```
wblock: Should this be all caps?  Even if the button is, the label here is meant to explain what the icon does.

Mrt134: Yes, this is meant to be in all caps, and the label is correct. If you log in to one of the testing systems and look at the option in question, it is identical to how the replacement is configured.

userguide/jails.rst
```diff
@@ -11,8 +11,8 @@ Jails
 
 Jails are a lightweight, operating-system-level virtualization.
 One or multiple services can run in a jail, isolating those services
-from the host %brand% system. %brand% uses the
-`iocage <https://github.com/iocage/iocage>`__ utility for jail
+from the host %brand% system. Beginning with %brand% 11.0, the
```
wblock: I don't think this is needed, and just adds details that will confuse the user.
```suggestion
from the host %brand% system. %brand% uses
```

userguide/jails.rst
```diff
@@ -11,8 +11,8 @@ Jails
...
+`iocage <https://github.com/iocage/iocage>`__ utility is used for jail
```
```suggestion
`iocage <https://github.com/iocage/iocage>`__ for jail
```

userguide/jails.rst
```diff
@@ -11,8 +11,8 @@ Jails
...
 management. Jails are also used as the basis for %brand% :ref:`Plugins`.
```
```suggestion
management. Jails are themselves used as the basis for %brand% :ref:`Plugins`.
```

PR #1578, opened 2019-10-03
NAS-103321 doc system>general> language field changes

a-stjohn:


userguide/system.rst
```diff
@@ -130,7 +130,8 @@ settings.
    |                      |                | maximum age to *31536000* seconds (one year). This means that after a browser connects to the %brand%                    |
    |                      |                | |web-ui| for the first time, the browser continues to use HTTPS and renews this setting every year.                      |
    +----------------------+----------------+--------------------------------------------------------------------------------------------------------------------------+
-   | Language             | combo box      | Select a language by typing in the field or selecting it from the dop-down menu. View the status of a language in the    |
+   | Language             | combo box      | Select a language from the drop-down menu. The list can be sorted by :guilabel:`Name` or :guilabel:`Language code`.      |
```
wblock: Language code should be a link to a reference on the ISO 639-1 codes, like here: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

PR #1582, opened 2019-10-04
Cherry-picked the docker doc building to angulargui.

a-stjohn:
Most credit goes to Kris since he helped debug and fix the
problems in the Makefile.


Mrt134:
I checked with Kris and this has been overtaken by events, so it should be closed.

PR #1599, opened 2019-11-05
NAS-101310: Update replication "State" to describe realtime updates

Mrt134:
HTML build test: no issues.


userguide/tasks.rst
```diff
@@ -1132,10 +1132,9 @@ method is selected.
 Saving a new replication adds an entry to
 :menuselection:`Tasks --> Replication Tasks`.
 The columns show the various settings for the replication. The
-:guilabel:`State` shows if the replication has run successfully or if
-an error has occurred. The log for the finished replication task can
-be viewed and downloaded by clicking the entry in the
-:guilabel:`State` column.
+:guilabel:`State` shows the result of the most recent replication and
+updates when a new replication starts. Clicking a :guilabel:`State`
+entry opens the log for the current or most recent replication.
```
wblock: "most recent" works for both of these.
```suggestion
entry opens the log for the most recent replication.
```

userguide/tasks.rst
```diff
@@ -934,11 +934,10 @@ that dataset is also created.
...
 :menuselection:`Tasks --> Replication Tasks`.
 The task settings that are shown by default can be adjusted by opening
-the :guilabel:`COLUMNS` drop-down. To see more details about the last
-time the replication task ran, click the entry under the
-:guilabel:`State` column. Tasks can also be expanded by clicking
-|ui-chevron-right| for that task. Expanded tasks show all replication
-settings and have |ui-task-run|, |ui-edit|, and |ui-delete| buttons.
+the :guilabel:`COLUMNS` drop-down. Tasks can also be expanded by
+clicking |ui-chevron-right| for that task. Expanded tasks show all
+replication settings and have |ui-task-run|, |ui-edit|, and |ui-delete|
```
wblock: We need to work toward mentioning capabilities generally rather than using lists of specific buttons or menu entries.  In this case, I'm not even sure this sentence is needed, particularly if a screenshot shows an expanded task.  If we don't have that screenshot, we should.

userguide/tasks.rst
```diff
@@ -1132,10 +1131,8 @@ method is selected.
 Saving a new replication adds an entry to
 :menuselection:`Tasks --> Replication Tasks`.
 The columns show the various settings for the replication. The
-:guilabel:`State` shows if the replication has run successfully or if
-an error has occurred. The log for the finished replication task can
-be viewed and downloaded by clicking the entry in the
-:guilabel:`State` column.
+:guilabel:`State` shows the status of the most recent replication.
+Clicking a :guilabel:`State` shows the replication log.
```
wblock: This is not always true, though.  Some states do not have logs.

PR #1605, opened 2019-11-12
NAS-103536: Document plugin startup behavior

Mrt134:
- Also clean up some extra whitespaces.
- HTML build test: no issues.


userguide/plugins.rst
```diff
@@ -101,16 +101,19 @@ with :guilabel:`NAT`. It is recommended to keep these plugins set to
...
 
 Click :guilabel:`ADVANCED PLUGIN INSTALLATION` to show all options for the
 plugin jail. The options are described in :ref:`Advanced Jail Creation`.
 
 To start the installation, click :guilabel:`SAVE`.
 
-Depending upon the size of the application, the installation may take a few minutes
-to download and install. A confirmation message displays once the plugin installation
-completes, along with any post-installation notes. 
+Depending upon the size of the application, the plugin can take several
```
wblock:
```suggestion
Plugin download and installation can take several minutes, depending on
```

userguide/plugins.rst
```diff
@@ -101,16 +101,19 @@ with :guilabel:`NAT`. It is recommended to keep these plugins set to
...
+minutes to download and install. A confirmation message displays when
```
wblock:
```suggestion
the size of the application. A confirmation message displays when
```

userguide/plugins.rst
```diff
@@ -101,16 +101,19 @@ with :guilabel:`NAT`. It is recommended to keep these plugins set to
...
+the plugin installation completes, along with any post-installation
+notes. Installed plugins are immediately activated and set to start when
```
wblock: What does "immediately activated" mean?  Started after install?  If so, this should be a separate sentence.  The next two sentences can be about how they default to autostart, but can be changed.

userguide/plugins.rst
```diff
@@ -130,17 +133,16 @@ page as shown in :numref:`Figure %s <view_installed_plugins_fig>`.
    Viewing Installed Plugins
 
 
-In addition to the :guilabel:`Jail` name, the :guilabel:`Columns` menu controls
-which information is displayed for each installed Plugins entry. Options include
-status, IPv4 address, IPv6 address, installed version, plugin name, release, boot status,
-and collection.
+In addition to the :guilabel:`Jail` name, the :guilabel:`Columns` menu
+controls which information is displayed for each installed Plugins
+entry. Options include status, IPv4 address, IPv6 address, installed
```
wblock: Let's not list default columns, which can change at any time. The user can see which columns can be displayed in the Columns menu.
```suggestion
entry.
```

userguide/plugins.rst
```diff
@@ -130,17 +133,16 @@ page as shown in :numref:`Figure %s <view_installed_plugins_fig>`.
...
+version, plugin name, release, boot status, and collection.
```
wblock:
```suggestion
```

userguide/plugins.rst
```diff
@@ -101,16 +101,20 @@ with :guilabel:`NAT`. It is recommended to keep these plugins set to
...
-Depending upon the size of the application, the installation may take a few minutes
-to download and install. A confirmation message displays once the plugin installation
-completes, along with any post-installation notes. 
+Plugin download and installation can take several minutes, depending on
+the size of the application. A confirmation message displays when the
+plugin installation completes, along with any post-installation notes.
+
+Installed plugins are immediately started. By default, all plugins are
+started when the system boots. Unsetting :guilabel:`Boot` means the
+plugin must be started manually.
```
wblock:
```suggestion
plugin will not start when the %brand% system boots, and must be started manually when needed.
```

userguide/plugins.rst
```diff
@@ -130,17 +134,15 @@ page as shown in :numref:`Figure %s <view_installed_plugins_fig>`.
...
+controls which information is displayed for each installed Plugins
+entry.
 
-More information such as *RELEASE*, *AUTOSTART*, and
-*VERSION* is shown by clicking |ui-chevron-right|. Options to
-:guilabel:`RESTART`, :guilabel:`STOP`, :guilabel:`UPDATE`,
-:guilabel:`MANAGE`, and :guilabel:`UNINSTALL` the plugin are also
-displayed. If an installed plugin has notes, the notes can be viewed by
-clicking :guilabel:`POST INSTALL NOTES`.
+More information such as *RELEASE* and *VERSION* is shown by clicking
```
wblock:
```suggestion
More information like *RELEASE* and *VERSION* can be shown by clicking
```

userguide/plugins.rst
```diff
@@ -130,17 +134,15 @@ page as shown in :numref:`Figure %s <view_installed_plugins_fig>`.
...
+|ui-chevron-right|. Options to :guilabel:`RESTART`, :guilabel:`STOP`,
```
wblock: Avoid making lists of specific action buttons or menu entries. These are on the screen, and having static lists in the doc just means it will be out of sync when the program changes.  In this case, it would be enough to say that there are buttons to control and manage the plugin.

Mrt134: This text has existed in angulargui for roughly 4-5 months (https://github.com/freenas/freenas-docs/blame/angulargui/userguide/plugins.rst), so this alteration is outside the scope of this ticket and PR. Please make a Jira ticket or Epic with the specific rework request so this concept can be prioritized then applied uniformly to the entire guide, rather than playing "whack a mole" ad infinitum.

userguide/plugins.rst
wblock: The changed version is somewhat redundant ("started" and "will start").  It's not clear what the difference is there.  "when the system boots" is unspecific as to which system is being talked about. The last sentence uses negative logic and an unspecific "requires": what requires what?  Is it a dependency, or something the system does?  The suggestion avoids most of these problems and is shorter.
```
A plugin is started immediately after installation and will start when the system boots. To disable boot startup and require manually starting the plugin, unset :guilabel:`Boot`.
```

PR #1610, opened 2019-11-13
Fix docs for editing interface.

a-stjohn:


dlavigne:
Why the change? Both of these fields are read-only for Edit.

a-stjohn:
> Why the change? Both of these fields are read-only for Edit.

When editing an interface the "Type" field is not displayed. However, the "name" is displayed and is the only read-only field on the page.

dlavigne:
Exactly. The user still needs to know that they are stuck with the type they choose as they won't be able to edit it. If you remove Type from that sentence the user assumes they can edit anything except for Name.

a-stjohn:
So then what do I need to change based on [this comment](https://jira.ixsystems.com/browse/NAS-101830?focusedCommentId=101693&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-101693)?

dlavigne:
Nothing, this PR should just be closed.

PR #1612, opened 2019-11-14
NAS-103528: Clean up remaining mentions of Graphite.

Mrt134:
- Remove obsolete sentence in reporting.rst
- Reapply original field name and description for System/Reporting `Graphite Server` option.
- Requires syncing with webui repo.
- HTML build test: no issues.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -1103,8 +1103,8 @@ These settings are described in
    | Report CPU usage    | checkbox  | Report CPU usage in percent instead of units of     |
    | in percent          |           | kernel time.                                        |
    +---------------------+-----------+-----------------------------------------------------+
-   | Graphite Server     | string    | Destination hostname or IP address for collectd     |
-   |                     |           | data sent by the Graphite plugin.                   |
+   | Remote Graphite     | string    | IP address or hostname of a remote server running   |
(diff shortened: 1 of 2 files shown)


userguide/system.rst
```diff
@@ -1103,8 +1103,8 @@ These settings are described in
    | Report CPU usage    | checkbox  | Report CPU usage in percent instead of units of     |
    | in percent          |           | kernel time.                                        |
    +---------------------+-----------+-----------------------------------------------------+
-   | Graphite Server     | string    | Destination hostname or IP address for collectd     |
-   |                     |           | data sent by the Graphite plugin.                   |
+   | Remote Graphite     | string    | IP address or hostname of a remote server running   |
```
wblock: We should prefer hostname to IP address, and mention it first.
```suggestion
   | Remote Graphite     | string    | Hostname or IP address of a remote    |
```

userguide/system.rst
```diff
@@ -1103,8 +1103,8 @@ These settings are described in
...
+   | Server Hostname     |           | `Graphite <http://graphiteapp.org/>`__.             |
```
wblock: Clearer to just say "Graphite server".
```suggestion
   | Server Hostname     |           | `Graphite <http://graphiteapp.org/>`__ server.             |
```

PR #1623, opened 2019-11-26
NAS-103773: Review and update encryption options for Cloud Sync Tasks.

Mrt134:
- Retest the behavior and update the Remote Encryption, Filename Encryption, Encryption Password, and Encryption Salt descriptions.
- Sync final text with UI help text.
- HTML build test: no issues.

### DIFF
--- a/userguide/tasks.rst
+++ b/userguide/tasks.rst
@@ -1676,8 +1676,8 @@ shows the configuration options for Cloud Syncs.
    +---------------------+----------------+------------------------------------------------------------------------------------------------------------+
    | Post-script         | string         | A script to execute after the Cloud Sync Task is run.                                                      |
    +---------------------+----------------+------------------------------------------------------------------------------------------------------------+
-   | Remote Encryption   | checkbox       | Use `rclone Crypt <https://rclone.org/crypt/>`__ to manage data encryption during *PUSH* or *PULL*         |
-   |                     |                | transfers:                                                                                                 |
+   | Remote Encryption   | checkbox       | Use `rclone crypt <https://rclone.org/crypt/>`__ to manage data encryption                                 |
+   |                     |                | during *PUSH* or *PULL* transfers:                                                                         |
    |                     |                |                                                                                                            |
    |                     |                | *PUSH:* Encrypt files before transfer and store the encrypted files on the remote system. Files are        |
    |                     |                | encrypted using the :guilabel:`Encryption Password` and :guilabel:`Encryption Salt` values.                |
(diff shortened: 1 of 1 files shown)


userguide/tasks.rst
```diff
@@ -1676,18 +1676,36 @@ shows the configuration options for Cloud Syncs.
...
+   |                     |                | that was used to encrypt the files.                                                                        |
+   |                     |                |                                                                                                            |
+   |                     |                | Adds the :guilabel:`Filename Encryption`, :guilabel:`Encryption Password`, and :guilabel:`Encryption Salt` |
+   |                     |                | options.                                                                                                   |
    +---------------------+----------------+------------------------------------------------------------------------------------------------------------+
-   | Filename Encryption | checkbox       | Encrypt the shared file names. Only appears when :guilabel:`Remote encryption` is enabled.                 |
+   | Filename Encryption | checkbox       | Include the filename with the :guilabel:`Remote Encryption` operation:                                     |
```
themylogin: > Include the filename with the :guilabel:`Remote Encryption` operation:

Very unclear for me what it means. I propose something like

Encrypt (PUSH) or decrypt (PULL) filenames. Filenames will be kept in secret, but directory structure will be preserved and identical files names will have identical uploaded names. Please see https://rclone.org/crypt/#file-name-encryption-modes for more details on filename encryption algorithm (we use "standard" algorithm).

userguide/tasks.rst
```diff
@@ -1676,18 +1676,36 @@ shows the configuration options for Cloud Syncs.
...
+   |                     |                |                                                                                                            |
+   |                     |                | *PUSH:* Encrypt the file name in addition to the file contents.                                            |
+   |                     |                |                                                                                                            |
+   |                     |                | *PULL:* Decrypt the file name when the file is transferred from the remote system. Decrypting the file     |
```
themylogin: The cause of this very ticket was to say here:

"When pulling with filename decryption and specifying incorrect password or salt, no files will be pulled but operation will be considered successful. "

userguide/tasks.rst
```diff
@@ -1676,18 +1676,36 @@ shows the configuration options for Cloud Syncs.
...
+   |                     |                |                                                                                                            |
+   |                     |                | *PULL:* Decrypt files that are being stored on the remote system before the transfer. Transferring the     |
+   |                     |                | encrypted files requires entering the same :guilabel:`Encryption Password` and :guilabel:`Encryption Salt` |
+   |                     |                | that was used to encrypt the files.                                                                        |
+   |                     |                |                                                                                                            |
+   |                     |                | Adds the :guilabel:`Filename Encryption`, :guilabel:`Encryption Password`, and :guilabel:`Encryption Salt` |
+   |                     |                | options.                                                                                                   |
```
themylogin: I think it would be fair and useful to add this link: https://rclone.org/crypt/#file-formats so users can understand what key derivation and encryption algorithms are used.

userguide/tasks.rst
Mrt134: On this point - Is this style of message (success, but nothing happened) immutable or is there some idea that this will be changed?

themylogin: It will look like the remote was empty and there is nothing we can do about it.

Mrt134:
Updated with review comments addressed!

PR #1628, opened 2019-12-02
DOCS-78: Add text about ACL inheritance

Mrt134:
- Adapted from https://www.reddit.com/r/freenas/comments/dy3unu/how_to_have_the_permissions_of_new_files/f87makn/?context=3
- HTML build test: no issues.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -975,7 +975,7 @@ configure the system to always display advanced settings by enabling the
 
 
 After a dataset is created it appears in
-:menuselection:`Storage --> Pools.`
+:menuselection:`Storage --> Pools`.
 Click |ui-options| on an existing dataset to configure these options:
 
 .. _storage dataset options:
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -1295,6 +1295,23 @@ These permissions control the actions users can perform on the dataset
 contents. ACLs are typically used to manage user interactions with
 :ref:`shared datasets <Sharing>`.
 
+The ACL for a new file or directory is typically determined by the
+parent directory ACL. An exception is when there are no *File* or
```
wblock: I think this should be `*File Inherit*` and the next sentence factored out the repeated word. We should include it for clarity.

userguide/storage.rst
```diff
@@ -1295,6 +1295,23 @@ These permissions control the actions users can perform on the dataset
...
+*Directory Inherit* :ref:`flags <ACE Inheritance Flags>` in the parent
+ACL :literal:`owner@`, :literal:`group@`, or :literal:`everyone@`
+entries. In this case the non-inheriting entries for these principals
```
wblock: In which case?  This is non-specific.  This is talking about the exception, but is not clear about it.

PR #1640, opened 2019-12-16
NAS-104007 update docs for new iocage pool activation dialog.

a-stjohn:
**11.3-RC1** change.

### DIFF
--- a/userguide/jails.rst
+++ b/userguide/jails.rst
@@ -52,7 +52,7 @@ intended jails and plugins. The
 screen displays a message and button to :guilabel:`CREATE POOL` if no
 pools exist on the %brand% system.
 
-If a pool exist, but no pool has been activated with iocage, a
+If pools exist, but none have been chosen for use with jails or plugins, a
 drop-down appears to select a pool. Select an appropriate pool and
 click :guilabel:`CHOOSE`.
 
--- a/userguide/jails.rst
+++ b/userguide/jails.rst
@@ -59,7 +59,7 @@ click :guilabel:`CHOOSE`.
 Multiple pools can be activated to store iocage jails and plugins. To
 select another pool, click |ui-settings|. A drop-down shows the
 active pool. Another pool can be selected from the drop down to
-to be used for iocage jails and plugins.
+to be used for jails and plugins.
 
 Jails and downloaded FreeBSD release files are stored in a dataset named
 :file:`iocage/`.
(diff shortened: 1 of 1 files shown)


userguide/jails.rst
```diff
@@ -52,15 +52,14 @@ intended jails and plugins. The
...
-Multiple pools can be activated to store iocage jails and plugins. After
-a pool is created, the
-:menuselection:`Jails`
-page displays an :guilabel:`Activated Pool` section. This shows which
-pool and iocage dataset is active with %brand%. Click :guilabel:`CONFIG`
-to view the option to choose another pool or dataset to activate with
-iocage. :guilabel:`ACTIVATE` another pool to refresh the
-:menuselection:`Jails`
-list with any jails that exist on the chosen pool or dataset.
+If a pool exist, but no pool has been activated with iocage, a
```
wblock: This can be more general.
```suggestion
If pools exist, but none have been chosen for use with jails or plugins, a
```

userguide/jails.rst
```diff
@@ -52,15 +52,14 @@ intended jails and plugins. The
...
+drop-down appears to select a pool. Select an appropriate pool and
```
wblock: Drop-down is an implementation detail, what the user sees is a dialog.
The user doesn't know which pools are "appropriate", so saying that just adds stress to the choice.
```suggestion
dialog appears to choose a pool. Select a pool and
```

userguide/jails.rst
```diff
@@ -52,15 +52,14 @@ intended jails and plugins. The
...
+click :guilabel:`CHOOSE`.
+
+Multiple pools can be activated to store iocage jails and plugins. To
```
wblock: Not clear what this means.  It sounds like I can choose more than one pool for jails and plugins, but that's not what the UI does.  It just lets you choose a different one.  It's not necessary to refer to iocage here at all, another implementation detail.

userguide/jails.rst
```diff
@@ -52,15 +52,14 @@ intended jails and plugins. The
...
+select another pool, click |ui-settings|. A drop-down shows the
+active pool. Another pool can be selected from the drop down to
```
wblock: "Another", or "A different"?

userguide/jails.rst
```diff
@@ -52,15 +52,14 @@ intended jails and plugins. The
...
+to be used for iocage jails and plugins.
```
```suggestion
to be used for jails and plugins.
```

PR #1651, opened 2019-12-16
NAS-103956 add instructions for reset data/passwords for SED.

a-stjohn:


userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
        LockOnReset:     1
 
 
+.. index:: SED, SED Password
+.. _Managing SED Password and Data:
+
+Managing SED Password and Data
```
```suggestion
Managing SED Passwords and Data
```

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This section contains command line instructions to manage SED
+passwords and data. The command used is
+`sedutil-cli(8) <https://www.mankier.com/8/sedutil-cli>`__. Most
+SEDs are TCG-E (Enterprise) or TCG-Opal
```
wblock: Not clear why this sentence is here. As the reader, why do I need to know this?  *Do* I need to know this?

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+(`Opal v2.0 <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Opal_SSC_v2.01_rev1.00.pdf>`__).
+
+.. warning:: These commands can be descructive to data and passwords.
```
```suggestion
.. warning:: These commands can be destructive to data and passwords.
```

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+   Take necessary data backup procedures and use the commands with
```
```suggestion
   Keep backups and use the commands with
```

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+   caution.
+
+Drives can be checked individually to determine if it is a SED and what
```
wblock: Too many things in one sentence.  Split out the part about individual or batch:
`Drive SED version can be checked on a single drive with this command:`
`All connected disks (just SED?) can be checked at once:`

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+variant of SED it is with this comand:
```
wblock: Typo.

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+
+.. code-block:: none
+
+   root@tn01b:~ # sedutil-cli --isValidSED /dev/da0
```
wblock: Probably clearer to replace the prompt with just `root@truenas`.

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+   /dev/da3 E Micron_5N/A U402
+   /dev/da4 E Micron_5N/A U402
+   /dev/da5 E Micron_5N/A U402
+   /dev/da6 E Micron_5N/A U402
+   /dev/da9 E Micron_5N/A U402
+   No more disks present ending scan
+   root@tn01b:~ #
+
+
+In the commands above, "E" represents that it is a TCG-E (Enterprise)
```
wblock: This would be easier to read either as a table.
 Table formatting will have to be fixed, since
Github isn't helping.
```
The value in the second column of output indicates the type of SED:

+-------------------------------------+
| :literal:`E`  | TCG-E (Enterprise) |
| :literal:`2`  | TCG-Opal (Opal v2.0) |
| :literal:`No` | Not a SED |
+-------------------------------------+
```

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+SED. If the drive is TCG-Opal (Opal v2.0), it has a "2" next to it. A
+drive that is not a SED has a "No" next to it.
+
+If the drive is **TCG-Opal**:
```
wblock: This should be a section to make it clear that all of the TCG-Opal commands are together.  Should TCG-E come first because it's more likely?

userguide/system.rst
```diff
@@ -822,6 +818,108 @@ with locking enabled:
...
+Wipe data and reset password to default MSID:
+:samp:`sedutil-cli --revertPer {oldpassword} /dev/{device}`
+
+Wipe data and reset password using the PSID:
+:samp:`sedutil-cli --yesIreallywanttoERASEALLmydatausingthePSID {PSINODASHED} /dev/{device}`
+where *PSINODASHED* is the PSID located on the pysical drive with no
+dashes (:literal:`-`).
+
+
+If the drive is **TCG-E**:
```
wblock: TCG-E section.

userguide/system.rst
a-stjohn: I would say they need to know this because there are different instructions for each.

wblock: Okay, so it should explain that after that sentence.
`Commands are different for the different drive types, so the first step is identifying which type is being used.`

userguide/system.rst
```diff
@@ -815,6 +811,111 @@ with locking enabled:
...
+SEDs are TCG-E (Enterprise) or TCG-Opal
+(`Opal v2.0 <https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage-Opal_SSC_v2.01_rev1.00.pdf>`__).
+Commands are different for the different drive types, so the first
+step is identifying which type is being used.
+
+.. warning:: These commands can be destructive to data and passwords.
+   Keep backups and use the commands with
+   caution.
+
+Drive SED version can be checked on a single drive with this command:
```
wblock: This can be like the second version:
```suggestion
Check SED version on a single drive, :literal:`/dev/da0` in this example:
```

userguide/system.rst
```diff
@@ -815,6 +811,111 @@ with locking enabled:
...
+   /dev/da2 E Micron_5N/A U402
+   /dev/da3 E Micron_5N/A U402
+   /dev/da4 E Micron_5N/A U402
+   /dev/da5 E Micron_5N/A U402
+   /dev/da6 E Micron_5N/A U402
+   /dev/da9 E Micron_5N/A U402
+   No more disks present ending scan
+   root@truenas:~ #
+
+
```
wblock: Did something happen to the explanation of what the output means?

PR #1662, opened 2019-12-18
NAS-104119 Add docs for "NAT Interface" field for jails.

a-stjohn:


userguide/jails.rst
```diff
@@ -569,77 +569,71 @@ Click :guilabel:`NEXT` to view all jail
...
    +------------------------+--------------+---------------------------------------------------------------------------------------------------------+
    | vnet3_mac              | string       | Leave this blank to generate random MAC addresses for the host and jail. To assign fixed MAC            |
    |                        |              | addresses, enter the host MAC address and the jail MAC address separated by a space.                    |
-   |                        |              |                                                                                                         |
+   +------------------------+--------------+---------------------------------------------------------------------------------------------------------+
+   | NAT Interface          | checkbox     | System network interface that remaps jail network addresses into a different IP address                 |
+   |                        |              | space. To override using the default system network interface, enter the name of a network              |
```
wblock: The second sentence is really wordy.  Shorter:
`Select a different interface to override the default.`

a-stjohn: You don't select an interface. It's a string only field. This is the help text which was already approved.

PR #1663, opened 2019-12-18
NAS-104321: Update ACL management text:

Mrt134:
- Add text about the directory/file browser identifying datasets that have an ACL.
- Add text about being required to use the ACL manager for dataset permissions.
- Remove obsolete warning about top-level datasets (these can no longer have their permissions changed in the web ui).
HTML build test: no issues.

### DIFF
--- a/userguide/storage.rst
+++ b/userguide/storage.rst
@@ -1321,7 +1321,7 @@ ACLs are modified by adding or removing Access Control Entries (ACEs) in
 :menuselection:`Storage --> Pools`.
 Find the desired dataset, click |ui-options|, and select
 :guilabel:`Edit ACL`. The :guilabel:`ACL Manager` opens. The ACL manager
-is required to modify permissions on a dataset that has an ACL.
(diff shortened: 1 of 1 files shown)


userguide/storage.rst
```diff
@@ -1317,10 +1318,8 @@ Datasets optimized for SMB sharing can restrict ACL changes. See
 ACLs are modified by adding or removing Access Control Entries (ACEs) in
 :menuselection:`Storage --> Pools`.
 Find the desired dataset, click |ui-options|, and select
-:guilabel:`Edit ACL`. The :guilabel:`ACL Manager` opens.
-
-.. warning:: Editing top-level datasets can prevent users from
-   accessing data in child datasets.
+:guilabel:`Edit ACL`. The :guilabel:`ACL Manager` opens. The ACL manager
+is required to modify permissions on a dataset that has an ACL.
```
wblock: Ambiguous "required".
```suggestion
must be used to modify permissions on a dataset with an ACL.
```

PR #1671, opened 2019-12-20
NAS-103887 Clarify docs on jail config backup location.

a-stjohn:
**11.3-RC1** change.

### DIFF
--- a/userguide/jails.rst
+++ b/userguide/jails.rst
@@ -93,7 +93,8 @@ Notes about the :file:`iocage/` dataset:
 * :file:`iocage/` datasets on activated pools are independent of each
   other and do **not** share any data.
 
-.. note:: The iocage config is backed up to
+.. note:: iocage jail configs are stored in
+   :file:`/mnt/{pool}/iocage/jails/{jail}`. These are backed up to
    :ref:`new boot environments <Boot>`. This means that rolling back
    to a previous boot environment preserves jail and plugin settings.
 
--- a/userguide/jails.rst
+++ b/userguide/jails.rst
@@ -94,7 +94,7 @@ Notes about the :file:`iocage/` dataset:
   other and do **not** share any data.
 
 .. note:: iocage jail configs are stored in
-   :file:`/mnt/{pool}/iocage/jails/{jail}`. These are backed up to
+   :file:`/mnt/{poolname}/iocage/jails/{jailname}`. These are backed up to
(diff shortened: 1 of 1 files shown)


userguide/jails.rst
```diff
@@ -93,7 +93,8 @@ Notes about the :file:`iocage/` dataset:
 * :file:`iocage/` datasets on activated pools are independent of each
   other and do **not** share any data.
 
-.. note:: The iocage config is backed up to
+.. note:: iocage jail configs are stored in
+   :file:`/mnt/{pool}/iocage/jails/{jail}`. These are backed up to
```
```suggestion
   :file:`/mnt/{poolname}/iocage/jails/{jailname}`. These are backed up to
```

userguide/jails.rst
```diff
@@ -93,7 +93,8 @@ Notes about the :file:`iocage/` dataset:
 * :file:`iocage/` datasets on activated pools are independent of each
   other and do **not** share any data.
 
-.. note:: The iocage config is backed up to
+.. note:: iocage jail configs are stored in
+   :file:`/mnt/{poolname}/iocage/jails/{jailname}`. These are backed up to
```
sonicaj: This is accurate, however they are not backed up to any BE. Not sure i follow that ?

a-stjohn: So it is accurate enough to remove that bit and just say 
```
iocage jail configs are stored in
:file:`/mnt/{poolname}/iocage/jails/{jailname}`. This means that rolling back
to a previous boot environment preserves jail and plugin settings.
```

sonicaj: So that is accurate but we should clarify. The configs aren’t stored in BE, rolling back BE would leave us with working iocage jails, however the configs would have to be manually restored.

a-stjohn: ahh okay. Could you explain how to manually restore the config so I can properly document it for users?

sonicaj: @a-stjohn so each time iocage updates the configuration file, existing configuration file ( `/mnt/{poolname}/iocage/jails/{jailname}/config.json`) is backed up like this `/mnt/{poolname}/iocage/jails/{jailname}/config_backup.json`. In order for the user to restore this, he/she needs to change the filename of `config_backup.json` to `config.json` ( preferably backing up `config.json` itself before doing this but of course it's up to them. If they intend to not save it, so be it, on an older version, it won't be usable anyways

a-stjohn: Awesome! Thank you!

PR #1675, opened 2019-12-26
DOCS-116: Update iSCSI Portal IP Addresses descriptions.

Mrt134:
- Add text about IPv6 wildcards.
- HTML/TrueNAS HTML build test: no issues.

### DIFF
--- a/userguide/sharing.rst
+++ b/userguide/sharing.rst
@@ -1499,12 +1499,10 @@ To configure iSCSI, click :guilabel:`WIZARD` and follow each step:
      This is required when the :guilabel:`Discovery Auth Method` is set
      to *CHAP* or *Mutual CHAP*.
 
-   * :guilabel:`IP`: Select the IP addresses associated with an
-     interface or the wildcard addresses :literal:`0.0.0.0` (all IPv4
-     addresses) and/or :literal:`::` (all IPv6 addresses). Choosing a
-     wildcard address prevents selecting individual IP addresses that
-     are the same type as the wildcard. To add more IP addresses on a
-     different :guilabel:`Port`, click :guilabel:`Add`.
+   * :guilabel:`IP`: Select IP addresses to be listened on by the portal.
+     Click :guilabel:`ADD` to add IP addresses with a different network
+     port. The address :literal:`0.0.0.0` can be selected to listen on
+     all IPv4 addresses, or :literal:`::` to listen on all IPv6 addresses.
 
    * :guilabel:`Port`: TCP port used to access the iSCSI target.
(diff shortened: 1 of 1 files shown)


userguide/sharing.rst
```diff
@@ -1637,10 +1638,10 @@ summarizes the settings that can be configured when adding a portal.
...
    |                       |           |                                                                             |
    +-----------------------+-----------+-----------------------------------------------------------------------------+
-   | IP address            | drop-down | Select the IPv4 or IPv6 address associated with an interface or the         |
-   |                       | menu      | wildcard address of *0.0.0.0* (any interface). Click :guilabel:`ADD` to add |
-   |                       |           | another IP address to the portal. Click the associated :guilabel:`DELETE`   |
-   |                       |           | button to remove the extra IP address.                                      |
+   | IP address            | drop-down | Select the IPv4 or IPv6 addresses associated with an interface or the       |
+   |                       | menu      | wildcard addresses :literal:`0.0.0.0` (all IPv4 addresses) and/or           |
+   |                       |           | :literal:`::` (all IPv6 addresses). To add more IP addresses on another     |
+   |                       |           | :guilabel:`Port`, click :guilabel:`ADD`.                                    |
```
wblock: I think there is also an value for "all addresses", although I'm not sure what was chosen.  If any of the "all" values are picked, it disables choosing other individual addresses, so the last sentence ("To add more") should become a second sentence, and the "all" aliases mentioned in a third sentence.

Mrt134: Just rechecked, there is no "all addresses" option, just the 0.0.0.0 and :: wildcards.

userguide/sharing.rst
```diff
@@ -1499,8 +1499,12 @@ To configure iSCSI, click :guilabel:`WIZARD` and follow each step:
      This is required when the :guilabel:`Discovery Auth Method` is set
      to *CHAP* or *Mutual CHAP*.
 
-   * :guilabel:`IP`: Select the IP address associated with an
-     interface or the wildcard address of *0.0.0.0* (any interface).
+   * :guilabel:`IP`: Select the IP addresses associated with an
```
wblock: The iSCSI Portals page doesn't even show network interfaces, so that reference is not needed.
Other conditions are essentially error handling, and don't need to be mentioned.
```
Select IP addresses to be listened on by the portal. Click :guilabel:`ADD
to add IP addresses with a different network port. The address
:literal:`0.0.0.0` can be entered to listen on all IPv4 addresses, or
:literal:`::` to listen on all IPv6 addresses.
```

PR #1677, opened 2019-12-26
DOCS-115 Update system>reporting docs.

a-stjohn:


userguide/system.rst
```diff
@@ -1165,12 +1165,11 @@ These settings are described in
...
    +-----------------------+-----------+-----------------------------------------------------+
-   | Confirm RRD Destroy   | checkbox  | Destroy the reporting database. Appears when        |
-   |                       |           | :guilabel:`Graph Age` or :guilabel:`Graph Points`   |
-   |                       |           | are changed. Required for changes to                |
-   |                       |           | :guilabel:`Graph Age` or :guilabel:`Graph Points`   |
-   |                       |           | to take effect.                                     |
-   +-----------------------+-----------+-----------------------------------------------------+
+
+Any changes made to :ref:`Reporting settings <reporting_options>`
+requires the report history to be cleared. A dialog is shown when a
```
wblock: The first sentence does not make it clear that the user gets a chance to cancel the change, even though the second sentence implies it.  How about this (not great, but clearer):
```
Changes to :ref:`Reporting settings <reporting_options>`
clear the report history. To keep history with the old 
settings, cancel the warning dialog.
```

userguide/system.rst
```diff
@@ -1165,12 +1165,11 @@ These settings are described in
...
+setting is changed. :ref:`Reporting settings <reporting_options>` can
+be reset to default values by clicking :guilabel:`RESET TO DEFAULTS`.
```
wblock: This last sentence is a little redundant.
```
Click :guilabel:`RESET TO DEFAULTS` to restore the original settings.
```

PR #1690, opened 2020-01-06
DOCS-180 add docs for button to API docs in UI.

a-stjohn:


a-stjohn:
> Looks good - one random thought: is it possible or beneficial to add a link to the FreeNAS websocket API docs here too?

Probably not. The URL for the docs uses whatever hostname/IP address the system is at. It just adds `/api/docs` to the end of the system URL.

Mrt134:
Understood, no worries then,

PR #1693, opened 2020-01-08
DOCS-97: TrueNAS Updates and Trains rework

Mrt134:
- Update this section to reflect the feature release and bugfix update behavior.
- Generalize the text about Production Use trains and add the 11.2 and 11.3 trains.
- TrueNAS HTMl build test: no issues.


Mrt134:
Got Kris' approval offline.

PR #1731, opened 2020-03-13
DOCS-353 Update jail upgrading docs.

a-stjohn:


a-stjohn:
> Are jail OS upgrades not handled in the UI?

I did check the UI. I didn't see any means for _upgrading_. There is an option to update them through the UI, though.

PR #1740, opened 2020-04-03
DOCS-193: Create single TrueNAS guide

Mrt134:
Remove ifdefs
Unify text files
Adjust images
Adjust identical reference paths
Add TrueNAS Enterprise callouts where appropriate
Fix all build issues
Update conf.py for correct version and remove all tag system code
Remove many obsolete files


userguide/storage.rst
```diff
@@ -23,13 +23,12 @@ these options:
 * :ref:`Multipaths`: View multipath information for systems with
   compatible hardware.
 
-#ifdef truenas
-.. note:: When using an HA (High Availability) %brand% system,
+
+.. note:: When using a |enterprise| HA (High Availability) %brand% system,
```
q5sys, line 27: Isn't an article still needed here? If I'm undertsanding the substitution right, this would read:
`When using TrueNAS Enterprise HA (High Availablility) TrueNAS system,`
Do we need an `a` or do we need to s/ `systems`

Mrt134: forgot to remove the extra %brand%, it should read:
"When using a TrueNAS Enterprise HA (High Availability) system,"

Mrt134, line 27:
```suggestion
.. note:: When using a |enterprise| HA (High Availability) system,
```

PR #1757, opened 2020-04-08
DOCS-123: review hardware recommendations and make greater conformity…

Mrt134:
… to official recommendations

- Review the larger chapter and make a number of corrections.

### DIFF
--- a/userguide/intro.rst
+++ b/userguide/intro.rst
@@ -273,13 +273,13 @@ Using the |Web-UI|
 Hardware Recommendations
 ------------------------
 
-%brand% |release| is based on FreeBSD 11.3 and supports the same
+%brand% |release| is based on FreeBSD 12.1 and supports the same
 hardware found in the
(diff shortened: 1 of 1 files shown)


userguide/intro.rst
```diff
@@ -288,22 +288,19 @@ called *amd64* by AMD and *Intel 64* by Intel.
    firmware interface or EFI.
 
 Actual hardware requirements vary depending on the workflow of your
-%brand% system. This section provides some starter guidelines. The
-`FreeNAS® Hardware Forum
+%brand% system. This section provides some starter guidelines.
+`The Official FreeNAS® Hardware Guide
```
q5sys, line 292: Since this is for v12... should this be TrueNAS Hardware Guide?  I assume we're going to be renaming the older one

Mrt134: I could rename it here, but the link will open a page that is still titled "FreeNAS Hardware guide" and I don't know if that post will be renamed...

PR #1775, opened 2020-05-11
DOCS-426: port docs improvements to 12.0 guide

Mrt134:
cherry-pick commits and rework where needed.

### DIFF
--- a/userguide/system.rst
+++ b/userguide/system.rst
@@ -1274,7 +1274,7 @@ has options to configure each %brand% :ref:`Alert`.
 
 
 Alerts are grouped by |web-ui| feature or service monitor. To customize alert
-importantance, use the :guilabel:`Warning Level` drop-down. To adjust how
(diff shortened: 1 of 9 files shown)


userguide/system.rst
```diff
@@ -1273,9 +1273,11 @@ has options to configure each %brand% :ref:`Alert`.
    Alert Settings
 
 
-Alerts are grouped by |web-ui| feature or service monitor. To adjust the
-alert level or notification frequency, choose an option from the
-drop-down menus and click :guilabel:`SAVE`.
+Alerts are grouped by |web-ui| feature or service monitor. To customize how
+important to flag an alert, use the :guilabel:`Warning Level` drop-down. To
```
q5sys, line 1277:
```suggestion
important an alert is flagged, use the :guilabel:`Warning Level` drop-down. To
```

q5sys: Idk if you'll like the uses of a past participle, so if you have another idea that's fine with me.

PR #1782, opened 2020-05-21
DOCS-106099 host allow and deny doc improvement.

a-stjohn:


userguide/sharing.rst
```diff
@@ -203,16 +203,31 @@ information given when the share was created.
...
    | Auxiliary Parameters         | string        | ✓        | Enter any additional `afp.conf <https://www.freebsd.org/cgi/man.cgi?query=afp.conf>`__ parameters             |
    |                              |               |          | not covered by other option fields.                                                                           |
    |                              |               |          |                                                                                                               |
    +------------------------------+---------------+----------+---------------------------------------------------------------------------------------------------------------+
 
+.. note::
+   :name: hosts allow/deny note
+
+   If neither *Hosts Allow* or *Hosts Deny* contains an entry, then allow
+   access.
```
Mrt134, line 220: This reads like it's missing objects? How about something like ... "then AFP share access is allowed for any host."   ?

PR #1784, opened 2020-05-26
11.3 angular - periodic snapshot task documentation changes

ianmtaylor1:
Fixed outdated information and added more details in Periodic Snapshot Tasks section.

* Changed default naming schema to current value.
* Removed sentence stating that snapshot lifetime is automatically appended to name (not true in version 11.3)
* Added subsection about how snapshots are automatically removed and a related warning for modifying or deleting periodic snapshot tasks.

This information is true to the best of my knowledge and has been tested on a fresh Virtualbox installation of FreeNAS 11.3-U3.1.

### DIFF
--- a/userguide/tasks.rst
+++ b/userguide/tasks.rst
@@ -818,6 +818,57 @@ not permitted while that replication task remains active. The
 replication task must be disabled before the related periodic snapshot
 task can be deleted.
 
+.. _Snapshot Autoremoval:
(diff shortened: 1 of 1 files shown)


a-stjohn:
Awesome! Thank you so much for contributing! At first glance, it appears to look like good work! Requesting review from appropriate team members. 😄

